PR #42, opened 2019-06-24
added scraper for Czech VR Network

crwxaj:


cld9x:
Haven't tried that yet, but looks solid!

I was meant to check myself if CzechVR networks sites follow same HTML template, as it might be possible to scrape them all with same code - maybe you had a chance to look at it?

cld9x:
Scratch that - I didn't realised you're scraping CzechVR network site with all the vids

cld9x:
Works great, thanks for the contrib!

PR #106, opened 2019-08-31
fix missing thrift host, use local modules

razfag:
Apache thrift dependency can't be fetched since git.apache.org is dead, it's now on github.
Also I guess using the rename command in go.mod would enable to use local packags.


cld9x:
Thanks for the effort, but this is caused by current outage of `git.apache.org` which can be simply solved by using go module proxy.

CI builds got hit by that outage just yesterday and https://github.com/cld9x/xbvr/commit/7361f3575565e888aaea0844cbfd1cc614a4905d has helped without going through too many hoops!

PR #114, opened 2019-09-19
Add HoloGirlsVR LethalHardcoreVR WhorecraftVR scrapers

jrebey:
I created a HoloGirlsVR scraper and tested it with the latest version. There are no filenames in the parsed HTML so I left that out like you did in other scrapers.

### DIFF
--- a/pkg/scrape/lethalhardcorevr.go
+++ b/pkg/scrape/lethalhardcorevr.go
@@ -110,8 +110,10 @@ func ScrapeLethalHardcoreVR(knownScenes []string, out *[]ScrapedScene) error {
         k := strings.Split(e.Attr("content"), ",")
         for i, tag := range k {
                 if i == len(k)-1 {
-                        if funk.ContainsString(sc.Cast, tag) {
-                                tag = strings.Replace(tag, sc.Cast[0], "", -1)
+        for _, actor := range sc.Cast {
+                          if funk.Contains(tag, actor) {
+                              tag = strings.Replace(tag, actor, "", -1)
+                }
                         }
                 }
                 tag = strings.ToLower(strings.TrimSpace(tag))
--- a/pkg/scrape/lethalhardcorevr.go
+++ b/pkg/scrape/lethalhardcorevr.go
@@ -1,142 +1,142 @@
 package scrape
 
 import (
-  "log"
-  "regexp"
-  "strings"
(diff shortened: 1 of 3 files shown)


cld9x:
Nice!

jrebey:
I just added another scraper :)

pkg/scrape/lethalhardcorevr.go
```diff
@@ -0,0 +1,138 @@
...
+                sc.Cast = append(sc.Cast, strings.TrimSpace(e.Attr("title")))
+        }
+      })
+
+      // Tags
+      e.ForEach(`meta[name=Keywords]`, func(id int, e *colly.HTMLElement) {
+        k := strings.Split(e.Attr("content"), ",")
+        for i, tag := range k {
+                if i == len(k)-1 {
+                        tag = strings.Replace(tag, sc.Cast[0], "", -1)
```
cld9x: Crashing here if no cast data gets collected

pkg/scrape/lethalhardcorevr.go
```diff
@@ -0,0 +1,138 @@
...
+  "log"
+  "regexp"
+  "strings"
+
+  "github.com/gocolly/colly"
+  "github.com/mozillazg/go-slugify"
+  "github.com/thoas/go-funk"
+)
+
+func isGoodTag(lookup string) bool {
```
cld9x: This is interesting approach - there's no prefilter in any other scraper, but it probably should be!

jrebey: I added it because the quality of the tags is horrible... and they're really HTTP meta keywords for SEO with scene tags stuffed in.

This site appears to be affiliated with adultempire somehow and the videos are iframes to adultempire.com. There was actually a little more useful data in regular DOM tags in the iframe but for simplicity I decided to parse the data from the scene page instead.

pkg/scrape/lethalhardcorevr.go
jrebey: I'm not getting a crash here. I'm still running the 1.x version though.. could that be the difference?

Do you have a specific visited URL I can take a look at?

jrebey: I found the issue and attached the commit to this pull request.

PR #115, opened 2019-09-28
Add RealityLovers and RealJam VR scrapers

jrebey:
Another scraper done!

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,2 @@
+tasks:
+  - command: go get -u -v github.com/UnnoTed/fileb0x github.com/cortesi/modd/cmd/modd
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 ## Features
 
 - Automatically match title, tags, cast, cover image, and more to your videos
-- Support for all the most popular VR sites: BadoinkVR, CzechVR Network, DDFNetworkVR, MilfVR, NaughtyAmericaVR, SexBabesVR, StasyQVR, TmwVRnet, VirtualRealPorn, VirtualTaboo, VRBangers, VRHush, VRLatina and WankzVR
+- Support for all the most popular VR sites: BadoinkVR, CzechVR Network, DDFNetworkVR, HologirlsVR, LethalHardcoreVR, MilfVR, NaughtyAmericaVR, SexBabesVR, StasyQVR, TmwVRnet, VirtualRealPorn, VirtualTaboo, VRBangers, VRHush, VRLatina and WankzVR
 - Supports DeoVR web mode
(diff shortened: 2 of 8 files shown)


jrebey:
ugh... I thought I was working on RealJam VR in another branch... but, sadly, it was in master so it got attached here as well. Updated pull request description.

jrebey:
I'm deleting this pull request. I found a bug in colly that is preventing some scenes from being scraped... not sure why, but sceneCollector.OnHTML(`html`... isn't finding anything on some pages. I verified that they do indeed have <html> tags so I'm not sure what is going on.

jrebey:
Okay, I think I found the issue but I'm not sure how to proceed. realitylovers.com seems to be throwing errors when too many requests are made. I'm not sure how to either 1) throttle the requests or 2) determine when this happens, invalidate the cache for the given URL, and retry the request.

I could disable the cache in the scraper, which I did in the testing and it eventually scrapes all the scenes but it took about 5 runs to get all 284 videos scraped.

Do you have any ideas? I'll keep digging into colly docs to see if I can do #2 which seems like the preferred route.

jrebey:
I figured it out and added retry logic to the code. I've been able to successfully scrape all sites that throttle requests.

jrebey:
We should probably add the retry logic to all scrapers... I didn't want to take that liberty so I just did it in the ones I wrote.

cld9x:
Thanks for the effort so far, new additions looks solid!

I think best way forward is to stop scraping the site as soon as first errors appears - this won't be ideal for a number of reasons (e.g. lack of ability to report site errors in UI, and more), but the whole scraping part begs for refactoring, so I guess it could be addressed later.

Had a quick look on colly's `OnError` handler and I think it should handle non-2xx HTTP responses - if that's the case it should enable capturing terminated requests, and cancelling queued ones.

If you feel like guessing what throttling rules were implemented, `LimitRule` is there ;)

jrebey:
I implemented OnError in my latest commit. I only handle the HTTP 429 Too many requests error right now. All other errors are just ignored by colly which is the existing behavior in the scrapers. What I’d really like to do is figure out how to invalidate the cache on all errors so that we aren’t caching them.

I looked at the limit options. We aren’t using more than one thread... or at least it didn’t appear so. When I set the limit to 2 threads I hit a lot more 429 errors. The only thing we can do is add a delay to every request. I decided it was better to simply pause when we overload the server rather than slow down the already long scraping process. I could possibly read the retry-after header in the response (if it exists) and respect that if you’d like.

jrebey:
I’m going to abandon this pull request and rebase it with a cherry pick. I used gitpod and instead of working in a branch it committed to master. Plus it’s got every commit from the last pull request.

PR #128, opened 2019-10-06
Refactor scraping so db/export scraping remains in sync

jrebey:
So that we don't need to duplicate code in both places, I moved everything out to a new function. This is going to make the threading much easier.


jrebey:
#133 contains these commits... I'm closing this one.

PR #132, opened 2019-10-12
Add file creation time to files view

jrebey:


cld9x:
Guess that's to scratch some immediate itch?

Merging that now, but this part of UI could use some brainstorming

PR #133, opened 2019-10-13
Scraper threading

jrebey:
This is a pretty big update involving a bunch of refactoring to support packages being able to access things in other packages. There is probably a bit more refactoring that could be done (moving API stuff out of xbvr)... but I figured I'd stop here as it's all relevant to the threading/UI changes.

New features:

- Scrapers run in parallel if XBVR_THREADING environment variable is set. This results in a 6-10x improvement in scrape time without increasing the load on each site from the previous version. When scenes are cached, updates are measured in seconds now :)
- Behind the scenes, scrapers use channels to communicate with writers
- The UI now displays an indeterminate progress bar for each scraper as they are running
- Added "Run All Scrapers" button to... run all scrapers... This ignores the enabled state of each scraper.


pkg/scrape/navr.go
```diff
@@ -23,6 +27,7 @@ func NaughtyAmericaVR(knownScenes []string, out *[]ScrapedScene) error {
 		colly.AllowedDomains("www.naughtyamerica.com"),
 		colly.CacheDir(sceneCacheDir),
 		colly.UserAgent(userAgent),
+		colly.Async(true),
```
cld9x: Why this one is async?

pkg/xbvr/task_content.go
```diff
@@ -122,51 +181,46 @@ func ScrapeJAVR(queryString string) {
...
 		}
 
 		content, err := json.MarshalIndent(out, "", " ")
 		if err == nil {
-			fName := filepath.Join(appDir, fmt.Sprintf("content-bundle-%v.json", time.Now().Unix()))
+			fName := filepath.Join(common.AppDir, fmt.Sprintf("content-bundle-%v.json", time.Now().Unix()))
 			err = ioutil.WriteFile(fName, content, 0644)
 			if err == nil {
-				tlog.Infof("Export complete, file saved to %v", fName)
+				tlog.Infof("Export completed in %v, file saved to %v", time.Now().Sub(t0), fName)
```
cld9x: Can we ISO-format the date here?

cld9x:
Indeed quite a lot of refactoring in this one, but it's all needed - moving models and db handling to separate package was long overdue!

Site options UI got overly busy IMO, but we can address that later.

Unless better config mechanism is in place, env vars are way to go - I think we should be recommending docker as the best way to test experimental features as it separates your live and test envs, plus setting variables is easy.

pkg/scrape/navr.go
jrebey: LOL.. cuz I had a ton of changes in flight and missed it ;)

I wondered why it completed so fast.

pkg/xbvr/task_content.go
jrebey: It's not a date.. it's a duration i.e. 3m30s.

PR #145, opened 2019-10-22
Update docker images and drone docker build options

jrebey:
Update docker images to use distroless containers and make drone use its own Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,42 @@
-FROM ubuntu:19.04
-RUN apt update && apt install -y wget ca-certificates
+FROM node:10.16.3-jessie as build-env
 
-ARG DRONE_TAG
+### Install Go ###
+ENV GO_VERSION=1.12.10 \
+    GOPATH=$HOME/go-packages \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,13 +1,42 @@
...
-    rm /tmp/xbvr.tgz
+WORKDIR /app
+ADD . /app
+RUN cd /app && \
+    yarn install && \
+    yarn build && \
+    go generate && \
+    go build -tags='json1' -ldflags '-w' -o xbvr main.go
+
+FROM FROM gcr.io/distroless/base-debian10
```
cld9x: `FROM FROM`? 😄 

Also, why debian10?

jrebey: grrr... I fixed that and then something must have reverted. I'll fix it.

jrebey: Google updated the distroless to debian10 but they pulled the images for some reason so I reverted it back to base.

jrebey:
FYI you can test a container built with this Dockerfile at jrebey/xbvr:latest

PR #175, opened 2019-10-31
Move Buefy styles to scss

jrebey:
This change allows us to easily modify the UI in the future. I've verified that all elements are the same using this method.

### DIFF
--- a/ui/src/main.js
+++ b/ui/src/main.js
@@ -6,7 +6,7 @@ import store from './store';
 import vueDebounce from 'vue-debounce';
 
 import Buefy from 'buefy';
-import 'buefy/dist/buefy.css';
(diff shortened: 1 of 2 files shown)


cld9x:
Is that preparation for future dark mode support?

If so, I'd like to hold it for a bit and investigate options properly

jrebey:
No. When I was playing with dark mode I noticed that it’s the recommended way to apply customization to Bulma/Buefy so I figured it would be good to do it now.

cld9x:
Sorry, but this breaks customizations that's applied by Buefy (which are part of `buefy/dist/buefy.css`) and few minor things is quite off.

Leaving it open until we have a good reason to start applying customizations to Buefy.

Issue #152, opened 2019-10-23
Star Ratings (1 - 5) for Scenes


vpfruitly:
Rather than just fav/not-fav, it would be nice to rate scenes on a scale.

jrebey:
This feature partially exists but is broken. There are ratings in Scene Details but setting a rating doesn't save nor is there a way to sort/filter on ratings.

cld9x:
Both saving and sorting by rating is in place already - the only thing missing is showing the rating in scene list

jrebey:
The rating also doesn’t display in the scene details. It shows no rating, so if it’s saving then it’s not pulling those fields in the view. Seems easy enough to fix.

PR #178, opened 2019-11-01
Fix issue with star ratings

jrebey:
Addresses one of the issues raised in #152 where after setting a scene rating, the rating doesn't persist in the UI.

### DIFF
--- a/ui/src/store/sceneList.js
+++ b/ui/src/store/sceneList.js
@@ -47,6 +47,14 @@ const mutations = {
       }
     });
   },
+  updateScene(state, payload) {
+    state.items = state.items.map(obj => {
+      if (obj.scene_id === payload.scene_id) {
+        obj = payload;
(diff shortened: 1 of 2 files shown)


cld9x:
This is caused by the fact that overlay is using scene data passed in props, so the right solution here is to either load scene data when showing overlay, or updating data in vuex store whenever change to it is made.

Reloading list with offset set to 0, will reset user position in the list, which is unwanted behaviour.

jrebey:
I think reloading when it’s changed is the right approach as it’ll minimize unnecessary io. I’ll make that change. 

This issue would prevent us from displaying the rating in the Scene list though until we have a method of anchoring the list (your url idea would help solve this).

cld9x:
Here's simplest take - scene object is updated directly inside the state, so no extra calls are neccesary

PR #179, opened 2019-11-02
Allow scene unmatching

### LINKED ISSUE #141: Ability to UNmatch and/or REmatch files to Scenes
Unless I'm being super dense and missing something, if you assign a file to a scene incorrectly, there's no way to fix that. Being able to unmatch or change the match would both solve the problem.

Perhaps linking the scene match modal right from the details view for a matched scene, and another button to just unmatch it entirely (for if you matched a file that has no corresponding scraper, to return it to bare file status).

jrebey:
Fixes #141

### DIFF
--- a/pkg/models/model_file.go
+++ b/pkg/models/model_file.go
@@ -40,6 +40,13 @@ func (f *File) Save() error {
 	return err
 }
 
+func (f *File) GetIfExist(id uint) error {
+	db, _ := GetDB()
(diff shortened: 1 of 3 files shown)


cld9x:
Had some issues with it, lets talk it through on discord

jrebey:
Closing as this revealed some major issues and requires much more work

PR #184, opened 2019-11-05
scraper for new VRBangers site

crwxaj:
VRBangers changed their site layout, using some AJAX requests with no proper HTML fallback, so I built this scraper. It works for me, but I'm new to Go, so the code style might not be the best, especially in regards to error handling. Please have a look.


cld9x:
Looks good to me!

Issue #169, opened 2019-10-28
Browse/Filter/Sort by File Metadata (Not Scene)


vpfruitly:
Hi, in closing #154 it made me think about how it would be useful to grab basic data on the files themselves. It looks like resolution and filesize is already captured, but framerate and bitrate (others?) would also be good.

At least once I have used those 3 values to determine which files to cull from my collection, so it may be of use to others in this way as well. I also used that info to figure out which scenes to upgrade, when I have had temporary access to a particular site. ie: Say you have a lot of 2160p files but get a new headset and want to replace those with 2700p versions.

Going a step further, being able to craft a filter that's like "site=Network && resolution < 4K" would be great, but that would require mapping all the various 4K-ish resolutions to a tag called "4K" (and ditto for tags like "5K" and "1600p" or whatever else. So I'd consider this a step beyond the basic level of displaying/filtering on raw values.

jrebey:
The data is already stored. We just need a good way of exposing the data in the UI in an unobtrusive way.

https://github.com/xbapps/xbvr/blob/master/pkg/models/model_file.go#L23-L29

cld9x:
I think such filter would make the most sense as a part of "files" UI, since in "scenes" view files are abstracted away.

cld9x:
Also, resolution and bitrate is already mentioned in scene details view!

vpfruitly:
lol, I knew I should have had it running and checked before typing out the whole ticket.

jrebey:
Do you still need this now that we realized the data is already there? In your use case, you can even remove the file in the same UI. Once I submit a small bug fix :)

vpfruitly:
Well, the file detail view doesn't expose bitrate, just resolution. But ideally it would be exposed in something like a table, rather than going into each file individually. I've got some thoughts on it, maybe it's something I could tackle when I get more familiar with the code.

vpfruitly:
I'd say this is done. Every filter I'd want is in there. Can probably generate new issues if anyone thinks of another filter they wanted.

PR #188, opened 2019-11-09
Files changes [WIP]

vpfruitly:
this is very much a work in progress. i'm hard coding english words into the vue files until we settle on what exactly needs to be used and what words, then it could get moved over the i8n json. none of the UI elements actually do anything. oh, i also changed "Match to scene" to just "Match" since we could really use that horizontal space in the files table, and i think two buttons named Play and Match look nicer, but you can be the final judge of that.


jrebey:
Fixes #169

vpfruitly:
Made some more changes, I'll get you a new PR soon.

Issue #142, opened 2019-10-22
Tracking Ticket for Requesting New Scrapers


vpfruitly:
I thought it might be handy to have a single issue to track requsted scrapers in one place. Here's all the ones requested so far from both here and https://feedback.xbvr.app/

- VR3000
- VirtualRealTrans
- VRBTrans
- VRConk
- SinsVR
- StockingsVR
- StripzVR
- GroobyVR
- VRP Films
- TSVirtualLovers

AlmopsVR:
Can you add https://sexlikereal.com ?

vpfruitly:
I don't think SLR is a studio that produces scenes, they're just a distributor. 

https://www.sexlikereal.com/virtualreality/studios

AlmopsVR:
True, but videos are hosted on the site and there's a lot of content.

vpfruitly:
SLR doesn't host all of the videos from each studio, they only get a selection, so the real solution is to scrape the studio directly. Look at individual videos on SLR and if their studio isn't already a scraper, or listed above, let us know. We can add that studio to the to-do list.

Sinoh726:
hello, can you please add https://wankitnowvr.com 
thank you!

dan9qt:
Might as well add zexyvr.com to go with wankitnowvr.com as they are from the same guys

SpedNSFW:
Two Naughty America companies:
 - https://www.realteensvr.com
 - https://www.randysroadstop.com

Others:
 - https://www.virtualxporn.com
 - https://www.vrclubz.com
 - https://www.littlecaprice-dreams.com/virtual-reality-little-caprice/#

Transcodes:
tonightsgirlfriend.com
vrfirsttimer.com

Aerowen:
@SpedNSFW I've added EvilEyeVR and VirtualXPorn to the SLR scraper. Even though VirtualXPorn does have its own site I found the scene details there are lacking a lot and they're added quickly on SLR so I'm sure that's fine for now.

LittleCapriceVR stopped releasing on SLR after three scenes that were mostly broken and needed fixing by SLR team, I'm not sure if the new ones on her own site are any better? (I don't feel like "renting" her scenes (lol), but that one may need a separate new scraper.

[...]

PR #191, opened 2019-11-10
Add VR3000 scraper

jrebey:
Another one for #142


jrebey:
If #199 is accepted, I'll need to modify this after it's merged. This doesn't need to block, the code will still work it'll just be using an older scraping method.

Issue #137, opened 2019-10-17
Not all filters are working on DLNA


dan9qt:
For example, no matter which site filter is chosen, all scenes are shown.

The actors filter on the other hand works as expected.

PR #198, opened 2019-11-11
Fix site and tag filter for DLNA (#137)

crwxaj:


cld9x:
Thanks a lot, I missed that

PR #199, opened 2019-11-12
Simplify scraper setup

jrebey:
In preparation for some documentation/tutorials I'm going to write for creating a new scraper, I figured I'd centralize the boilerplate code in every scraper. This has the added benefit of allowing us to update the OnError callback for retry/error handling logic across the board.

This should make scraper code easier to follow since there is a lot less code in some cases.

I also added ToJSON and Log methods to ScrapedScene which will be used in an example scraper to allow devs to easily print the output of the entire object instead of writing values to the database to see if things worked.


cld9x:
This is a good improvement, but you missed that site/scene collectors use different directories for cache

PR #205, opened 2019-11-13
Add WetVR scraper

crwxaj:
Added a scraper for wetvr.com. However, their CDN uses an URL format with meaningful double slashes, which doesn't work well with the image proxy. Not sure what I can do about that.


cld9x:
Looks like changes to image proxy code would be necessary (check `NewRequest` in imageproxy package)

jrebey:
This shouldn't require any modification to the proxy.

Are you talking about the link being:

```
cover-image="//sugmh11809khorp.belugacdn.link/0ihbyW_QasnJuDVMavPA8ae7qe4=/full-fit-in/970x545/center/middle/smart//content/videos/6/5/4/65495f8d-c248-409a-be20-a13be6d81dd3/handtouched/019.jpg"
```

That URL works if you visit:

https://sugmh11809khorp.belugacdn.link/0ihbyW_QasnJuDVMavPA8ae7qe4=/full-fit-in/970x545/center/middle/smart//content/videos/6/5/4/65495f8d-c248-409a-be20-a13be6d81dd3/handtouched/019.jpg

You're gonna need to add https: to that URL before adding it as a cover:

```
coverURL := e.ChildAttr(`div#t2019-video deo-video`, "cover-image")
sc.Covers = append(sc.Covers, "https:" + coverURL)
```

You'll have to do the same for gallery images:

```
galleryURL := e.Attr("src")
sc.Gallery = append(sc.Gallery, "https:" + galleryURL)
```

crwxaj:
https is not the problem. Somewhere the double slashes between smart//content get removed and the URL is not valid anymore. This seems to be the default behavior of http/Client, and I'm not sure where this can be disabled.

jrebey:
Ah, I see now. Okay, that’s interesting.

crwxaj:
Removing the double slashes is indeed a fixed behavior of DefaultServeMux:
https://golang.org/src/net/http/server.go?s=71874:71942#L2312  (see cleanPath())
To fix that we would have to change to whole different Router implementation. I opted for a hacky workaround instead, where I reinsert the double slash before sending the request to the imageproxy handler.

vpfruitly:
One thought I had, looking at this: the sugmh11809khorp part of the full domain looks randomly generated and liable to change at any time without notice. Maybe change the match to just belugacdn.link under the assumption that all their hosts are going to do the same thing with the double slashes.

Also, crwxaj consider joining the Discord https://discord.gg/wdCHXAG there's a development channel and is very collaborative and friendly. I wouldn't have been able to push a single commit without it, though you obviously have more experience than I do.

cld9x:
@crwxaj that's a smart workaround, but I'd prefer to keep scraper stuff together if possible

I don't mind riding on a diff router, looks like gorilla mux have just what's needed: https://godoc.org/github.com/gorilla/mux#Router.SkipClean

PR #212, opened 2019-11-16
fix date parser for VRBangers

crwxaj:
Some of the new scenes turned up in 2018.


jrebey:
Confirmed this correctly parses the date format.

Issue #213, opened 2019-11-17
Possibility to sort Scenes by scenes.created_at and scenes.updated_at


Sinoh726:
If possible I kindly ask you to implement following enhancement:

On Scenes page in dropdown menu "Sort by" to add two new possibilities:
Date scraped - from database table Scenes.created_at
Date modified - from database table scenes.updated_at

Reason:
currently when user is running scrape on multiple screapers he get's information on number of new scenes scraped but he does not get information on what were those scenes. This option would b e very convenient specially in cases when selecting new scraper sites.

Second (Data modified) would display latest scenes on which user added files so user would have possibility to double check if he added file to wrong scene and have latest added on top.

jrebey:
This is actually a bug for the created_at request.

The request for updated_at won’t do what you want. It would have to be sorted by the files table because the association happens there. I’m not sure how you’d do that.

PR #214, opened 2019-11-17
Fix scene sorting by Date Added

jrebey:
Raised in #213


cld9x:
This isn't a bug - `AddedDate` field stores creation date of file associated with scene

Issue #202, opened 2019-11-13
Lock the scene search popup size


dan9qt:
This can resize a bit depending on what is found in the search, would be useful if it was fixed somehow so the assign button is always in a consistent place.

vpfruitly:
This has been added. The modal is now aligned to the top (rather than center), so that the first assign button is always in the same location.

Issue #195, opened 2019-11-10
Pagination related issues in the Scene Match


vpfruitly:
I love the pagination. It's awesome and makes matching multiple files in a row so much nicer. But, it has a few minor issues:

- [x] Results aren't removed before switching to next file, leading to stale/incorrect results.
- [x] You aren't returned to page 1 of results after changing the text, or moving to the next file.
- [x] The modal should loop from the last file to the first file

PR #221, opened 2019-11-22
File modal fixes

vpfruitly:
This fixes the remaining two items in #195


vpfruitly:
This also fixes #202 by affixing the modal to the top, rather than centering it. that way the first assign button is always in the same position, rather than moving up and down depending on how many results are returned.

cld9x:
I don't love top aligned modal, but this is good enough for now

PR #222, opened 2019-11-22
Add authentication to UI and DeoVR

### LINKED ISSUE #204: Secure/password protected browser access.
I found that if I open up 9999 on my router I can access the web interface and watch videos on DeoVR remotely from myip:9999.

I presume this is a very unsafe thing to do, so I have closed up the port in my router, but I wonder if in the future this could be done safely.

jrebey:
UI auth is enabled via setting UI_USERNAME and UI_PASSWORD environment variables

DeoVR auth is enabled via DEO_USERNAME and DEO_PASSWORD environment variables

Fixes #204


snickers2k:
O_o 
i don't have this option
Bildschirmfoto 2021-12-16 um 11 32 26

PR #225, opened 2019-11-25
Redesign Match Modal [WIP]

vpfruitly:
This PR does the following to the Files view:
- The files table is now paginated. With over 400 files, not only were load times getting unreasonably long, it was also an eye strain to take in all at once and too much scrolling.
- A header ("Actions") has been put above the column with scene match and delete buttons.
- You can view the scene details modal for files with a scene match.
- The Scene button is green when it has been matched and will display the details modal. The Scene button is red when the file has not been matched and will display the match modal.
- The Delete button has been changed to purple, and word wrapping has been applied to the filename in the confirmation dialog to prevent scrolling.
- The match modal has been slightly redesigned. The VR player is now included, along with details on the file itself (size, bitrate, location, etc). The table of results remains unchanged.
- The separate VR player modal has been removed.
- A long (but not exhaustive, additions are welcome!) list of garbage strings are removed from the filename when searching. This includes things like "180x180", "samsung", "3dh", "oculus", and much more. In my experience, removing these manually was something I often had to do in order to get really good match results.

Note: 1 line of code had to be changed in the sceneList.js file. The "<" link for moving to the "previous scene" was always being displayed. This is because the value of i in the function was being returned not as 0, but -1. The conditional has been adjusted to return null for 0 and -1 both.


vpfruitly:
too old, too late

PR #230, opened 2019-11-27
Grooby scraper

vpfruitly:
Horrible website. Barely any info for all scenes. Some scenes have an additional VOD page with a little more info, but some don't, and some that do are still missing information.


vpfruitly:
For #142

PR #238, opened 2019-12-22
Add support for unmatched files in DeoVR

jrebey:


cld9x:
Good stuff, but file needs to be checked for availability (same as scene), so we won't be showing files that can't be played

jrebey:
Those aren’t fields in the File struct. Is that something we should add? The closest thing we have in current state is checking the volume availability.

cld9x:
Yes, this is what needs to be checked - volume needs to be available when it's local (it doesn't matter for cloud volumes)

jrebey:
I didn't see a way of doing this in gorm via sql (determine if it's local and only filter available volumes if local). If you know of a way to do it, please let me know.

cld9x:
It's good enough for now!

PR #257, opened 2020-02-28
Several changes and fixes for NAVR, VRHush, RealJamVR, TmwVRnet scrapers

Aerowen:
Minor changes and fixes, but please do check if it's all working well. I did test it in gitpod and it all worked fine there but wasn't able to get it to commit my changes, so had to manually edit the changes in my fork on github. No idea how to compile to an .exe with gitpod, I'm assuming that's just not possible. So I couldn't test it locally.

- Changed NaughtyAmericaVR horizontal and vertical covers to use a larger size.
- Fixed title and cover scraping for RealJamVR which wasn't working any more.
- Fixed TmwVRnet scraper, same issue as RealJamVR because deo player element changed.
- Changed VRHush cover and gallery images to use better, larger images.
- Added a fix for double white spaces in model names in VRHush scraper.

About the last point, it was adding duplicate model names to the database (see Adriana Chechik). But I wonder if basic trimming/sanitizing of model names shouldn't be done at a higher level. Like just before adding to the database so it would do that for all scrapers.

### DIFF
--- a/pkg/scrape/navr.go
+++ b/pkg/scrape/navr.go
@@ -70,11 +70,11 @@ func NaughtyAmericaVR(wg *sync.WaitGroup, updateSite bool, knownScenes []string,
 			base := strings.Split(strings.Replace(e.Attr("src"), "//", "", -1), "/")
 
 			base[8] = "horizontal"
-			base[9] = "1252x708c.jpg"
+			base[9] = "1182x777c.jpg"
 			sc.Covers = append(sc.Covers, "https://"+strings.Join(base, "/"))
 
 			base[8] = "vertical"
-			base[9] = "400x605c.jpg"
+			base[9] = "1182x1788c.jpg"
 			sc.Covers = append(sc.Covers, "https://"+strings.Join(base, "/"))
 		})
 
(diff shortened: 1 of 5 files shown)


Aerowen:
Well fuck. I just had to add a missing "regexp" import which I messed up by adding "regex", so I wanted to undo that commit but apparently there is no "undo" in git even though all tutorials say there should be, all 'revert' does is add more nonsense commits instead of actually removing or reverting to earlier commit. I thought that was the whole purpose of git. :(

I think it's finally corrected but I'm scared to do anything else now as any mistake just keeps adding onto it. Also it seems I can't really change my master branch while this pull request is pending as it automatically adds things here even though I didn't push request those commits, this makes no sense to me. If there's a clear manual that clearly explains how this is supposed to be done, please send me a link. Because I'd like to contribute a bit to this but when it gets messy like this I just can't. :/

cld9x:
Thanks for these, I’ll test and review asap

Aerowen:
vpfruitly explained on discord things have been busy, so I guess I'll add scrapers and scraper fixes to this until it can be reviewed.

I just added a SLR scraper with 11 of the more actively producing studios that don't have their own website to scrape and publish their scenes on SLR. More could be added later, tho some don't produce scenes any more. I used [gjson](http://github.com/tidwall/gjson) for working with json, hope that's okay.

PR #262, opened 2020-04-05
Various fixes and changes to multiple scrapers.

Aerowen:
### Fixes/Changes to scrapers:

SLR
- Added StripzVR studio to SLR scraper.
- Use the newly added large covers (mostly for new scenes, not all older scenes have a large cover).
- Fixed everything that got broken when SLR changed the json metadata and some html.
- Tags were removed from json metadata entirely, changed it to grab the tags from html instead.

BaDoinkVR
- Use large covers for all BaDoink-sites (improves VRCosplayX & BabeVR).
- Added tags for VRCosplayX scenes, as they should always have "Cosplay" and "Parody" tags (Other sites already use those).

MilfVR
- Use slightly larger gallery images similar to WankzVR.

LethalhardcoreVR / WhorecraftVR
- Minor fix for a couple edge-cases, not all actor names were removed from tags (Isabel Moon, Carolina Sweets).

NaughtyAmericaVR
- Unescaped actor names because it's pulled from js (example cast: April O'Neil & Belle O'Hara).

Realitylovers
- Use larger cover and gallery images.

SexBabesVR
- Fix for the issue of every other gallery image being a blank.

VirtualTaboo
- Fix release date format (thx mauroamore).

VR3000
- Added alternative cover, it's a vertical DVD-style cover so the original banner cover is still used for thumbnail view.

VRBangers
- Added alternative cover, it's larger but also much wider so the small cover is still used for thumbnail view.

VRConk
- Fix for using the actual gallery images instead of the tiny thumbnails.

VRLatina
- Fix for edge cases: check for empty cast element and a cast name with a "!" in it.
- Fix for missing cast names with rare multi-girl scenes (only one cast name is featured, others are added as tags).
- Change castname in tag checking to EqualFold for case-insensitive comparisons (lower case actor tags were not filtered out, example tag: nikol sparta).

### DIFF
--- a/pkg/scrape/badoink.go
+++ b/pkg/scrape/badoink.go
@@ -40,9 +40,9 @@ func BadoinkSite(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out 
 		})
 
 		// Cover URLs
-		e.ForEach(`div#videoPreviewContainer picture source`, func(id int, e *colly.HTMLElement) {
+		e.ForEach(`div#videoPreviewContainer picture img`, func(id int, e *colly.HTMLElement) {
 			if id == 0 {
-				sc.Covers = append(sc.Covers, strings.Split(e.Request.AbsoluteURL(e.Attr("srcset")), "?")[0])
+				sc.Covers = append(sc.Covers, strings.Split(e.Attr("src"), "?")[0])
 			}
 		})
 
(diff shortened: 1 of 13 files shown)


Aerowen:
Previously I fixed an issue with VRHush having double spaces in a few cast names which was causing duplicate cast entries and scenes not linking to the correct name. But I noticed that now that's fixed those scenes are added again with the correct name assigned but it leaves the old scene entry as well which stay linked to the bad double-spaced actor names. So there are a few duplicates in existing databases.

I assumed a 'force update' would update with the corrected cast names but it doesn't. People would have to use the "delete all scraped scenes" option on the VRHush scraper and do a fresh scrape to get rid of those duplicate scenes. Is it possible to add a migration for it? I had a look at migrations.go but did not want to mess with that myself. It would have to only delete the VRHush scenes of the cast names who have double spaces, afaik only these actors have the issue: (Adriana  Chechik, Nicolette  Shea and Antonia  Sainz) but a search for two spaces in the name field of the actors table lists them all as well.

With this pull requests there's a fix for NaughtyAmericaVR with names that have single quotes like Belle O'Hara that's pulled from js and didn't get unescaped, but it seems those scenes are getting updated with the corrected name just fine. I'm not sure why it didn't with VRHush. This fix for NAVR does leave the bad names in the actor table but with a zero scenes count so they're hidden.

PR #271, opened 2020-04-28
Make scraper logging thread safe

jrebey:
Sometimes scrapers try to write their status at the same time causing the following error:

```
(/go/appsrc/pkg/models/model_site.go:27) 
[2020-04-28 00:40:45]  database is locked 
time="2020-04-28T00:40:45Z" level=error msg="database is locked"
```

This causes the daemon to die. This fix addresses this by creating a channel to send log messages to and it processes them in order ensuring that only one message has a db lock at any given time.

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -259,5 +259,4 @@ func Migrate() {
 	}
 	common.Log.Printf("Migration did run successfully")
 
-	db.Close()
 }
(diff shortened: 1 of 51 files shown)


jrebey:
Closing this... this needs way more work. Database access is all over the place, the UI tries to access the DB while it's scanning and runs into locking issues nonstop. This started happening once I upgraded to 0.3.0-beta-13 and it's still not working in beta-16... I think I was on beta-9 before when it was working. I can't complete a scrape now.

PR #281, opened 2020-05-13
Add MySQL/MariaDB as a database option

jrebey:
This PR adds the ability to use other databases besides sqlite3. If no database is specified, sqlite3 is used as a default.

To enable mysql set the following environment variables (using your own values):

DB_TYPE=mysql
DB_HOST=localhost
DB_NAME=xbvr
DB_USER=xbvr
DB_PASSWORD=password

I don't have a postgres installation setup but I'll send a followup PR for postgres once I do.

This PR also includes some fixes to the way models are written to the database. Using pointers is the preferred way for gorm to call Save() and Delete(). There was also a bug in DeleteLock() which was exposed when enabling mysql where the table name was not being applied to the query and mysql freaks out.

And lastly, added a SQL_DEBUG env var to dump SQL queries to the logs for troubleshooting... this was invaluable in debugging some of the queries that were failing in mysql.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -6,13 +6,15 @@ import (
 )
 
 var (
-	DEBUG          = os.Getenv("DEBUG")
-	MYSQL_HOST     = os.Getenv("MYSQL_HOST")
-	MYSQL_DB       = os.Getenv("MYSQL_DB")
-	MYSQL_USER     = os.Getenv("MYSQL_USER")
-	MYSQL_PASSWORD = os.Getenv("MYSQL_PASSWORD")
-	SQLITE_PARAMS  = os.Getenv("SQLITE_PARAMS")
-	WsAddr         = "0.0.0.0:9998"
+	DEBUG         = os.Getenv("DEBUG")
+	SQL_DEBUG     = envToBool("SQL_DEBUG", false)
+	DB_TYPE       = getEnv("DB_TYPE", "sqlite3")
+	DB_HOST       = os.Getenv("DB_HOST")
+	DB_NAME       = os.Getenv("DB_NAME")
+	DB_USER       = os.Getenv("DB_USER")
+	DB_PASSWORD   = os.Getenv("DB_PASSWORD")
+	SQLITE_PARAMS = os.Getenv("SQLITE_PARAMS")
+	WsAddr        = "0.0.0.0:9998"
 )
 
 func envToBool(envVar string, defaultVal bool) bool {
@@ -26,4 +28,11 @@ func envToBool(envVar string, defaultVal bool) bool {
 	} else {
 		return defaultVal
 	}
+}
+
+func getEnv(key, fallback string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return fallback
 }
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -8,10 +8,12 @@ import (
 )
 
 var (
-	DEBUG        = os.Getenv("DEBUG")
-	SQL_DEBUG    = envToBool("SQL_DEBUG", false)
-	DATABASE_URL = getEnv("DATABASE_URL", fmt.Sprintf("sqlite:%v", filepath.Join(AppDir, "main.db")))
-	WsAddr       = "0.0.0.0:9998"
+	DEBUG             = os.Getenv("DEBUG")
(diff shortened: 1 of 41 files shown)


pkg/common/common.go
```diff
@@ -7,6 +7,12 @@ import (
 
 var (
 	DEBUG         = os.Getenv("DEBUG")
+	SQL_DEBUG     = envToBool("SQL_DEBUG", false)
+	DB_TYPE       = getEnv("DB_TYPE", "sqlite3")
+	DB_HOST       = os.Getenv("DB_HOST")
+	DB_NAME       = os.Getenv("DB_NAME")
+	DB_USER       = os.Getenv("DB_USER")
+	DB_PASSWORD   = os.Getenv("DB_PASSWORD")
```
cld9x, line 15: maybe it would be better to have single `DATABASE_URL` var and then parse it for internal use?

pkg/models/utils.go
```diff
@@ -32,8 +34,14 @@ func GetDMSData() DMSData {
...
-	tx.Select("strftime('%Y-%m', release_date) as release_date_text").
-		Group("strftime('%Y-%m', release_date)").Find(&scenes)
+	switch common.DB_TYPE {
+	case "mariadb", "mysql":
+		tx.Select("DATE_FORMAT(release_date, '%Y-%m') as release_date_text").
+			Group("DATE_FORMAT(release_date, '%Y-%m')").Find(&scenes)
+	case "sqlite3":
+		tx.Select("strftime('%Y-%m', release_date) as release_date_text").
+			Group("strftime('%Y-%m', release_date)").Find(&scenes)
+	}
```
cld9x, line 44: I'd drop that filter and replace it with proper between-dates filter (requires UI changes ofc)

pkg/xbvr/api_scenes.go
```diff
@@ -136,8 +138,14 @@ func (i SceneResource) getFilters(req *restful.Request, resp *restful.Response)
...
-	tx.Select("strftime('%Y-%m', release_date) as release_date_text").
-		Group("strftime('%Y-%m', release_date)").Find(&scenes)
+	switch common.DB_TYPE {
+	case "mariadb", "mysql":
+		tx.Select("DATE_FORMAT(release_date, '%Y-%m') as release_date_text").
+			Group("DATE_FORMAT(release_date, '%Y-%m')").Find(&scenes)
+	case "sqlite3":
+		tx.Select("strftime('%Y-%m', release_date) as release_date_text").
+			Group("strftime('%Y-%m', release_date)").Find(&scenes)
+	}
```
cld9x, line 148: as above

pkg/xbvr/task_volume.go
```diff
@@ -59,8 +60,14 @@ func RescanVolumes() {
...
-
-			err := db.Raw("select scenes.* from scenes, json_each(scenes.filenames_arr) where lower(json_each.value) = ? group by scenes.scene_id", strings.ToLower(path.Base(fn))).Scan(&scenes).Error
+			var query string
+			switch common.DB_TYPE {
+			case "mariadb", "mysql":
+				query = fmt.Sprintf("select scenes.* from scenes where JSON_CONTAINS(LOWER(scenes.filenames_arr), '\"%v\"') group by scenes.scene_id", strings.ReplaceAll(strings.ToLower(path.Base(fn)), "'", "''"))
+			case "sqlite3":
+				query = fmt.Sprintf("select scenes.* from scenes, json_each(scenes.filenames_arr) where lower(json_each.value) = '%v' group by scenes.scene_id", strings.ReplaceAll(strings.ToLower(path.Base(fn)), "'", "''"))
+			}
+			err := db.Raw(query).Scan(&scenes).Error
```
cld9x, line 70: If going with multi-db, we should drop JSON at db level and replace it with more universal approach, e.g. here we could select all possible filenames and match it using go

jrebey: I made it work with the existing JSON to not break existing setups

pkg/common/common.go
jrebey: Done

pkg/models/utils.go
jrebey: This would break DLNA release date filtering so I left it.

pkg/xbvr/api_scenes.go
jrebey: see DLNA note

PR #286, opened 2020-05-16
Wrap all DB writes in retry logic

jrebey:
Anything that writes to the DB is subject to locking errors. This PR addresses that.

### DIFF
--- a/pkg/models/db.go
+++ b/pkg/models/db.go
@@ -12,9 +12,6 @@ import (
 
 var log = &common.Log
 
-type DBModel interface {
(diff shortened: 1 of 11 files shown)


pkg/models/db.go
```diff
@@ -12,6 +12,29 @@ import (
 
 var log = &common.Log
 
+type DBModel interface {
```
cld9x, line 15: what's DBModel for?

n-i-x: Oops... leftover from a different approach.

PR #308, opened 2020-07-26
Adds RandysRoadStop & RealTeensVR scrapers

SpedNSFW:
Adds RandysRoadStop & RealTeensVR scrapers through scraping vrporn.com

Both RandysRoadStop & RealTeensVR contain little to no scene information on their respective sites and had to resort to using vrporn.

### DIFF
--- a/pkg/scrape/vrporn.go
+++ b/pkg/scrape/vrporn.go
@@ -1,0 +1,133 @@
+package scrape
+
+import (
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/gocolly/colly"
+	"github.com/mozillazg/go-slugify"
+	"github.com/thoas/go-funk"
+	"github.com/xbapps/xbvr/pkg/models"
+)
+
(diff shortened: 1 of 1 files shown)


Aerowen:
Can you add (VR Porn) to the studios just like with the SLR scraper, so it'll show up on the Options>Scrapers page it's a scraper that grabs from VRPorn. Like:
`registerScraper("realteensvr", "RealTeensVR (VRPorn)", ...`

For synopis it's doing a ForEach p-element and it only stores the last one. It seems they can have multiple p-elements in there and sometimes an empty one last, like with [this scene](https://vrporn.com/ariel-grace-fucking-real-horny-teen/) so in that case the synopsis will be empty. This worked for me to just grab all paragraphs:
`sc.Synopsis = strings.TrimSpace(e.ChildText(".entry-content.post-video-description"))`
(I know a lot of the e.ForEach() used in most scrapers are a bit unnecessary even when we just grab a single specific element. I used existing scrapers as example too and they all kinda have this.)

The RealTeensVR descriptions are sadly mostly advert junk text instead of an actual description, but the RandysRoadstop ones are fine. (Now we only need XBVR to start showing synopsis in the app too heh).

Two regex lines can be moved to the top on this one as well. Aside from these small things I didn't find any other problems with this scraper.

PR #309, opened 2020-07-26
Add Little Caprice Dreams scraper

SpedNSFW:
Adds a scraper for Little Caprice Dreams.

There have been problems with images but that's out of my control. Images get junk like `#700x0` appended to the tail end and causes XBVR to throw an error when transforming.

```
2020/07/26 18:11:33 error transforming image https://media.littlecaprice-dreams.com/wp-content/uploads/2019/03/cropped-lcd-heart-32x32.png#128x0: image: unknown format
2020/07/26 18:11:33      187.4982ms    403    GET  /img/128x/https:/media.littlecaprice-dreams.com/wp-content/uploads/2019/03/cropped-lcd-heart-32x32.png
```

### DIFF
--- a/pkg/scrape/littlecaprice.go
+++ b/pkg/scrape/littlecaprice.go
@@ -1,0 +1,104 @@
+package scrape
+
+import (
+	"math/rand"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/gocolly/colly"
+	"github.com/mozillazg/go-slugify"
+	"github.com/thoas/go-funk"
+	"github.com/xbapps/xbvr/pkg/models"
+)
+
+func LittleCaprice(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out chan<- models.ScrapedScene) error {
+	defer wg.Done()
+	scraperID := "littlecaprice"
+	siteID := "Little Caprice Dreams"
+	logScrapeStart(scraperID, siteID)
+
+	sceneCollector := createCollector("www.littlecaprice-dreams.com")
+	siteCollector := createCollector("www.littlecaprice-dreams.com")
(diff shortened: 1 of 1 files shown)


Aerowen:
I checked this one out, good to see more people do scrapers! Some notes:

You're generating a randomized number for SiteID but that makes the SceneID random as well, which should always be the same every scrape iirc. I suggest using the last part of the URL like `little-caprice-vr-fuck-me` from `https://www.littlecaprice-dreams.com/project/little-caprice-vr-fuck-me/`, although some of those get a little long, perhaps adding the release date instead might be better so it's unique but not random.

For the synopsis, it's missing the dot in the `.vid_desc` selector so it's not being stored in the database. I hate the "E: or English:" part they start descriptions with so I tried to split it by : to remove that junk but it's missing on some scenes so that causes panic errors, sigh. Nothing is easy with Golang imho, it's why I prefer making complicated regex...

For the covers I think would be better to use the one from the listing like before clicking to the scene page, those don't have the black part on the left. I know it's going to be a bit tricky because it's not on the scene page. I don't think there's a scraper which does this already, maybe ask in discord.

The image errors are weird, they probably have things set-up very strictly as all those images with errors give me a 403 if I manually copy/paste them in browser. It seems if you remove the 'media' sub-domain from it they will work correctly, I think most of them already are without the media subdomain and those work fine.
So maybe try to check and fix it for all URLs before adding them to the database, a regex would probably be easiest for that (btw regex compile lines can be put just above the sceneCollector.OnHTML() so it only needs to do it once not for every scene page being scraped, see vrallure.go for example. This reminds me I still have to fix that on the SLR scraper).

Got the same 403 error with the site logo for the scraper, I suggest using this one instead without the sub-domain which does seem to work normally: `https://www.littlecaprice-dreams.com/wp-content/uploads/2019/03/cropped-lcd-heart-180x180.png`

PR #310, opened 2020-07-26
Sort scene tags alphabetically

SpedNSFW:
Simply sorts the listed scene tags in an order that makes sense: alphabetically. Some sites already sort their tags alphabetically but most seem to have them arbitrarily listed which can be annoying and unpleasant.

Before:
Before

After:
After

Depends on #339 as of eb50be1


cld9x:
Tags are currently sorted by number of occurrences, which I think is a better default than alphabetical sort.

This one qualifies as a configurable user preference, preferably with a switch to choose between global and local count.

PR #313, opened 2020-07-27
add descriptions to scene details view

SpedNSFW:


cld9x:
The reason why I haven't added descriptions to scene details page yet is because I couldn't find good place for that.

I don't like your approach as it takes plenty of vertical space. Could you try putting it into tab, right after "watch history"?

SpedNSFW:
> The reason why I haven't added descriptions to scene details page yet is because I couldn't find good place for that.
> 
> I don't like your approach as it takes plenty of vertical space. Could you try putting it into tab, right after "watch history"?

Sure, that works.

PR #316, opened 2020-07-28
WetVR scraper fixes

Aerowen:
Now scrapes actual release date and a sceneID from the site, grabbing of scene duration is also improved. A migration was added to flush WetVR scenes due to new SceneIDs.

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -262,6 +262,24 @@ func Migrate() {
 				return tx.AutoMigrate(Scene{}).Error
 			},
 		},
+		{
+			// WetVR has changed scene numbering schema, so it needs to be flushed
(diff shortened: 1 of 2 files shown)


pkg/migrations/migrations.go
```diff
@@ -262,6 +262,24 @@ func Migrate() {
...
+				var scenes []models.Scene
+				db.Where("site = ?", "WetVR").Find(&scenes)
+
+				for _, obj := range scenes {
+					files, _ := obj.GetFiles()
+					for _, file := range files {
+						file.SceneID = 0
+						file.Save()
+					}
+				}
```
SpedNSFW, line 278: What's the point in this if you're just deleting the rows from the database?

PR #319, opened 2020-07-31
Fix Preview Generation

SpedNSFW:
`...\xbvr\video_preview\tmp\concat` becomes `.../xbvr/video_preview/tmp/concat`

Unescaped backslashes will cause problems in ffmpeg arguments so we convert the path to a Unix format (which Windows still supports). For Unix systems this change would do absolutely nothing.

The alternative would be `strings.replace(concatFile, "\\", "\\\\", -1)` which just looks messy and confusing at a glance.

### DIFF
--- a/pkg/xbvr/task_preview.go
+++ b/pkg/xbvr/task_preview.go
@@ -130,9 +130,9 @@ func renderPreview(inputFile string, destFile string, startTime int, snippetLeng
 		"-y",
 		"-f", "concat",
 		"-safe", "0",
-		"-i", concatFile,
+		"-i", filepath.ToSlash(concatFile),
 		"-c", "copy",
-		destFile,
+		filepath.ToSlash(destFile),
 	}
 	err = exec.Command(GetBinPath("ffmpeg"), cmd...).Run()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


SpedNSFW:
I've added a fix for preview generation of TB and Mono videos. I've tested it on the 6 TB and 2 Mono videos in my library and it works a charm.

cld9x:
Fix for preview generation on Windows looks fine, but could you implement mono/top-bottom/left-right detection based on actual video file dimensions (or existing scene info) rather than its filename?

See https://github.com/xbapps/xbvr/blob/master/pkg/xbvr/task_volume.go#L188

SpedNSFW:
> Fix for preview generation on Windows looks fine, but could you implement mono/top-bottom/left-right detection based on actual video file dimensions (or existing scene info) rather than its filename?
> 
> See https://github.com/xbapps/xbvr/blob/master/pkg/xbvr/task_volume.go#L188

I can't use file dimensions for mono videos as they're the same dimensions as LR videos. If you look at `emilybloom-10749` for example, it's a mono video with a resolution of 3840x1920 which is identical to that of 1920p LR videos.

Now I can use a dimension check for TB scenes but there's no way to differentiate between Mono and LR scenes. If you have any other suggestions I'm open to it. I'm still new to the codebase so I'm not sure what's available.

PR #321, opened 2020-07-31
fix wankzvr description

SpedNSFW:
Because the child element is included in `e.Text`, its doubling up the result. You only need to grab it once.


theRealKLH:
Also with milfvr (uses same code I believe)?

PR #333, opened 2020-08-04
Model Aliases draft

SpedNSFW:
This is a draft. This might not work on your system.

Things to do:

- [ ] Use path for JSON that will work on every system and in compiled executable
- [x] Move the reading of the JSON out of the `ConvertName` function so it only gets read once (not once per name)


theRealKLH:
superceded by #953 ?

Issue #276, opened 2020-05-10
Displaying Scenes per multiple pages instead of one endless page


Sinoh726:
If possible to implement same style as on "Files" page, to display scenes on multiple pages istead of one infinite page with "Load more" on bottom. Currently there are well over 7000 Scenes available on scrapers and it is harder to browse through older scenes in current setup.

Thank you!

PR #339, opened 2020-08-08
Add User Preferences options page

SpedNSFW:
Because I didn't just want to add a blank page with a random save button, I added support for a toggle for #310.

This page can be used for other settings such as (but not limited to):

 - Dark/Light CSS toggles
 - Scene page display toggles (#276)

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -24,9 +24,6 @@
 </script>
 
 <style>
-  .b-radio.radio {
-    margin-right: 0 !important;
-  }
   .navbar-pad {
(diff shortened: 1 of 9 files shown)


ui/src/App.vue
```diff
@@ -24,9 +24,6 @@
 </script>
 
 <style>
-  .b-radio.radio {
-    margin-right: 0 !important;
-  }
```
SpedNSFW, line 29: This didn't seem to be doing anything other than affect the radios added here so I removed it. I did confirm that the other radios still look the way they should.

PR #343, opened 2020-08-11
Scene Editing

SpedNSFW:
preview

Adds a modal window that can be opened from either the scene lists or scene details views. Will allow to edit scene details including but not limited to title, description, cast, and tags.

 - [x] Add front-end
 - [x] Add back-end

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -18,7 +18,8 @@ type Object struct {
 		Password string `default:"" json:"password"`
 	} `json:"security"`
 	Web struct {
-		TagSort  string `default:"by-tag-count" json:"tagSort"`
+		TagSort   string `default:"by-tag-count" json:"tagSort"`
+		SceneEdit bool   `default:"false" json:"sceneEdit"`
 	} `json:"web"`
 	Interfaces struct {
 		DLNA struct {
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -298,6 +298,12 @@ func Migrate() {
 				return db.Where("site = ?", "EvilEyeVR").Delete(&models.Scene{}).Error
 			},
 		},
+		{
+			ID: "0015-scene-edits",
+			Migrate: func(tx *gorm.DB) error {
+				return tx.AutoMigrate(&models.Action{}).Error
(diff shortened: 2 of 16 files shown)


SpedNSFW:
While I'm confident that everything is working without any flaws, I'd still recommend backing up your database while testing.

cld9x:
This looks good, few comments though:

- edit button in scene list view is cluttering the space, but I understand it might be handy if doing mass edits - can we hide it by default, and expose only if explicitly enabled in UI preferences?
- manual file match should be captured as a scene edit (as we are adding some filename as one of possible options)
- changes in tags/cast aren't being captured,
- suggestions for tags/cast would be nice,
- what about list of possible filenames in edit dialog?

SpedNSFW:
> This looks good, few comments though:
> 
> * edit button in scene list view is cluttering the space, but I understand it might be handy if doing mass edits - can we hide it by default, and expose only if explicitly enabled in UI preferences?
> * manual file match should be captured as a scene edit (as we are adding some filename as one of possible options)
> * changes in tags/cast aren't being captured,
> * suggestions for tags/cast would be nice,
> * what about list of possible filenames in edit dialog?

1. I've added an option to enable/disable the edit button on the SceneCard view. It will be disabled by default.
2. I'll look into this a bit later. While I had intended this to eventually happen, it's not currently on my mind.
3. Done now. I wasn't a fan of the implementation when I came up with it but it's the only thing I can really come up with.
4. Autocomplete suggestions have been added.
5. I also want to add a list for gallery/cover images so once I figure out an effective way to do that without clutter, I'll add filenames too.

cld9x:
Nice one. I'll take care of storing changes when matching files, and we address the remaining bits separately

SpedNSFW:
I currently don't have any validation on the images. If it's wanted, I can do some basic validation and make sure it ends in common image extensions (jpg, jpeg, png, gif)

cld9x:
I'd say we can live without validations for now - normally there should be no reason to edit these, so I'd assume user knows what he's doing

SpedNSFW:
Added a confirmation prompt when the editor is closed when changes have been made and nothing has been saved.

PR #344, opened 2020-08-11
Tranz vr

theRealKLH:


theRealKLH:
3rd commit removes padding of blank images on TranzVR scene gallery

PR #350, opened 2020-08-12
Graveyard Sites

### LINKED ISSUE #349: Scrape-Once Scrapers
There are a lot of studios that have stopped producing scenes. It would be good to probably separate them from the rest of the scrapers in some way and make it clear that they only need to be scraped once and be done.

SpedNSFW:
Closes #349 

There are a lot of VR studios than no longer produce scenes, whether the company no longer operates or the studio is publishing under a new name.

This PR separates these sites from the main list by placing them below in their own category I've called "Graveyard Sites" which is not final, I just couldn't think of anything better, but it gets the point across. I also remove the enabled switch for these as there's no point in scraping these sites repeatedly.


Sinoh726:
maybe better name would be Obsolete sites ?

cld9x:
think "defunct" sounds better, less grim 😄

pkg/migrations/migrations.go
```diff
@@ -280,6 +280,15 @@ func Migrate() {
 				return db.Where("site = ?", "WetVR").Delete(&models.Scene{}).Error
 			},
 		},
+		{
+			ID: "0014",
```
cld9x, line 284: can you tweak migration name to be in line with recent ones? (xxxx-brief-description)

PR #351, opened 2020-08-12
fix force-update and delete scenes

### LINKED ISSUE #334: Cannot delete SLR scenes through the UI
Selecting "Delete scraped scenes" in the UI for any scraper that uses SLR does not appear to work. At this stage I'm unsure of what the cause is.

SpedNSFW:
For sites where their name doesn't match what's been set in the database (VRPorn and SLR namely), they cannot be deleted or force updated as the query will find no matches.

This PR strips the scraper suffix from the site name as well as tidies up two site names where they also wouldn't match.

Closes #334


crwxaj:
supersede by #568

PR #354, opened 2020-08-13
Update vrporn.go

theRealKLH:
removed 'vr' from function call


SpedNSFW:
*cough* cld9x 558214f *cough*

PR #357, opened 2020-08-18
Scraper Table

SpedNSFW:
This would serve as a replacement to the current Scrapers page. This is just a first draft and I already have improvements in mind.

Note: This requires buefy v0.9.0 however this version breaks CSS functions (see: buefy/buefy#2752)

### DIFF
--- a/pkg/xbvr/api_options.go
+++ b/pkg/xbvr/api_options.go
@@ -167,8 +167,21 @@ func (i ConfigResource) listSites(req *restful.Request, resp *restful.Response) 
 	db, _ := models.GetDB()
 	defer db.Close()
 
-	var sites []models.Site
-	db.Order("name asc").Find(&sites)
+	type Site struct {
+		ID         string    `json:"id"`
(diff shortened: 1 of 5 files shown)


pkg/xbvr/api_options.go
```diff
@@ -167,8 +167,21 @@ func (i ConfigResource) listSites(req *restful.Request, resp *restful.Response)
...
+		IsEnabled  bool      `json:"is_enabled"`
+		LastUpdate time.Time `json:"last_update"`
+	}
+
+	var sites []Site
+	db.Table("sites").
+		Select("DISTINCT sites.id, sites.name, sites.avatar_url, sites.is_enabled, sites.last_update, scenes.studio").
+		Joins("INNER JOIN scenes ON sites.name=scenes.site").
+		Order("sites.name ASC").
+		Scan(&sites)
```
SpedNSFW, line 184: There's probably a better way of doing this but I didn't want to touch too much.

PR #365, opened 2020-09-17
Fix VirtualRealPorn Site ID scraping

CoolMintChocolate:
VRP no longer gives the video id as part in a `link[rel=shortLink]`, so scraping videos from them does not work currently. Instead I'm scraping it from the `body` classes.

### DIFF
--- a/pkg/scrape/virtualrealporn.go
+++ b/pkg/scrape/virtualrealporn.go
@@ -34,14 +34,13 @@ func VirtualRealPornSite(wg *sync.WaitGroup, updateSite bool, knownScenes []stri
 
 		var tmpCast []string
 
-		e.ForEach(`body`, func(id int, e *colly.HTMLElement) {
-			bodyClasses := strings.Split(e.Attr("class"), " ")
-			for _, c := range bodyClasses {
-				if strings.HasPrefix(c, "postid-") {
-					sc.SiteID = strings.Split(c, "-")[1]
-					sc.SceneID = slugify.Slugify(sc.Site) + "-" + sc.SiteID
-				}
-			}
+		e.ForEach(`script[id="deovr-js-extra"]`, func(id int, e *colly.HTMLElement) {
+			var jsonObj map[string]interface{}
+			jsonData := e.Text[strings.Index(e.Text, "{") : len(e.Text)-12]
+			json.Unmarshal([]byte(jsonData), &jsonObj)
+
+			sc.SiteID = jsonObj["post_id"].(string)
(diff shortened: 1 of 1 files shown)


pkg/scrape/virtualrealporn.go
```diff
@@ -34,10 +34,14 @@ func VirtualRealPornSite(wg *sync.WaitGroup, updateSite bool, knownScenes []stri
 
 		var tmpCast []string
 
-		// Scene ID - get from URL
-		e.ForEach(`link[rel=shortlink]`, func(id int, e *colly.HTMLElement) {
-			sc.SiteID = strings.Split(e.Attr("href"), "?p=")[1]
-			sc.SceneID = slugify.Slugify(sc.Site) + "-" + sc.SiteID
+		e.ForEach(`body`, func(id int, e *colly.HTMLElement) {
```
jrebey, line 37: There is only every one <body> in a valid HTML document. There's no need for a ForEach.

Also, I'm not a huge fan of this method of getting the id as it relies on a class name.

There's a better way:

favScript := e.ChildText(".add-to-favs > script")

That should leave you with favScript being "var stars_post = '121466';"

Just parse the post out of that string. You could also parse the deovr javascript which might even be more reliable as it's less likely to change formats vs something VRP controls.

jrebey (changes requested):

pkg/scrape/virtualrealporn.go
CoolMintChocolate: That's actually my first preference as well, however I couldn't think of an easy way to parse the Javascript variable. I know that `colly` does not execute Javascript so we can't get those variables out automatically, so we do have to parse them ourselves. Do you have suggestions? regexp is always an option but it's always the last thing I want to use (cue [xkcd](https://xkcd.com/1171/))

theRealKLH: How about...

```
		e.ForEach(`script[id="deovr-js-extra"]`, func(id int, e *colly.HTMLElement) {
			var jsonObj map[string]interface{}
			jsonData := e.Text[strings.Index(e.Text, "{") : len(e.Text)-12]
			json.Unmarshal([]byte(jsonData), &jsonObj)

			sc.SiteID = jsonObj["post_id"].(string)

```

CoolMintChocolate: Thanks for the guidance, I'll push some changes soon.

CoolMintChocolate:
Update from the review has been pushed. Feel free to take another look.

theRealKLH:
closes #366

PR #368, opened 2020-10-17
Trick DeoVR into playing hi-res files

crwxaj:


theRealKLH (commented):
This is a temporary workaround until DeoVR devs resolve within the application.

PR #389, opened 2020-12-01
Virtualrealporn name fix

theRealKLH:
Not sure if automatching ever worked, but this should fix. Also added VRAmateur.


theRealKLH:
no actual change to the Dockerfile. My local copy is higher version. reset to "Master" version.

Issue #386, opened 2020-11-18
correct Wankzvr MilfVR and TranzVR Filenames


Deflated-Criossant:
Correct filename for wankzvr MilfVR and TranzVR 

**wankzvr-wrong-house-right-hole-180_180x180_3dh_LR.mp4**

for all formats

needs fixing
remove scene ID and device format
![wvr](https://user-images.githubusercontent.com/16240121/99546578-8f616200-29dc-11eb-84ad-8af5093c6d7f.jpg)

Similarly for MVR and TVR

milfvr-<>-180_180x180_3dh_LR.mp4
tranzvr-<>-180_180x180_3dh_LR.mp4

theRealKLH:
do you know all the naming for different formats? Is it just the 3? or is the like 4k,5k, etc?

PR #395, opened 2020-12-17
Update wankz.go

theRealKLH:
add "site" to filename for automatching


theRealKLH:
missed adding site (wankzvr- milfvr- tranzvr-) to filename when fixing closed #386

PR #410, opened 2021-01-25
Fix scene detail scrolling and scraper menu clipping

### LINKED ISSUE #326: Problems with using prev/next buttons for scenes
In the scenes view when a scene card is open, using the prev/next buttons can cause problems.

1. The gallery does not get updated and can stay on the previous scene's gallery.

2. Changing what image is active in the gallery before changing scenes will change the active image on the next instead of resetting to the first image.

syph-nsfw:
Two small UI fixes:
 - Resolve the issue when viewing scene details then navigating to a scene with fewer thumbnails. Reactivity resolution order causes getImageURL() calls beyond the array bounds.
 - Allow the scraper menus to overflow outside the card boundaries, preventing clipping.

Fixes #326

### DIFF
--- a/ui/src/views/options/sections/OptionsSceneDataScrapers.vue
+++ b/ui/src/views/options/sections/OptionsSceneDataScrapers.vue
@@ -141,6 +141,7 @@
 
   .card {
     height: 100%;
+    overflow: visible;
   }
 
   .card-content {
--- a/ui/src/views/scenes/Details.vue
+++ b/ui/src/views/scenes/Details.vue
@@ -349,7 +349,11 @@
         }
       },
       getIndicatorURL(idx) {
-        return this.getImageURL(this.images[idx].url, "x40")
+        if (this.images[idx] !== undefined) {
+          return this.getImageURL(this.images[idx].url, "x40");
+        } else {
+          return "";
+        }
       },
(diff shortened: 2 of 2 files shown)


theRealKLH:
Tested and this works great. Minor "cosmetics"... the gallery position only resets to the first image if the next scene has less images than your current image position. Ex. current scene I am on image 7. If next scene only has 5 images, then it will reset to image 1 when I move to it. Otherwise it stays at the same position.. current 2, next scene 2. current 5, next 5.. etc.

syph-nsfw:
The gallery position doesn't reset to the first image, it simply wraps around when the index exceeds the number of images. One thing I did notice is the last image is shown both at the last index, and one past it. I think this is more of a Buefy bug.

Assuming two scenes, A with 10 images and B with three, going from A -> B works as follows;
| A index | B index |
|---------|---------|
|    0    |    0    |
|    1    |    1    |
|    2    |    2    |
|    3    |    3    |
|    4    |    4    |
|    5    |    5    |
|    6    |    **5**    |
|    7    |    0    |
|    8    |    1    |
|    9    |    2    |

The gallery position can be reset on each scene change with a watcher, but I assume the current behavior is intended. I like scrolling through scenes on the second image - skipping over the cover image.

Issue #401, opened 2021-01-13
Remove VRConk duplicates


Deflated-Criossant:
keep https://vrconk.com/christmas-vacation-80
has more tags 

delete https://vrconk.com/virtualreality/scene/id/80-christmas-vacation

![vrconk dupes](https://user-images.githubusercontent.com/16240121/104411773-08baa280-5591-11eb-9c64-251633b1062d.jpg)

PR #418, opened 2021-02-02
fix VRConk scraper

crwxaj:
This scraper produced wrong scene IDs for a while, using the first word of the scene name. So there probably will be duplicate scenes.


crwxaj:
Related to #401. This PR won't remove the duplicates though.

PR #419, opened 2021-02-06
Track watching activity from DeoVR. Broken since d863d5e.

hjs0007:
I noticed that my files were not flagged anymore as watched when watched from DeoVR portal. The bug were introduced in commit d863d5e. Not sure what was the idea behind the "do not track" feature that was defined in that commit.


cld9x:
Tracking scene "watch" status is now done via DeoVR "remote mode"

Since it needs to be explicitly enabled in DeoVR, we should probably give users way to get back to the old way via settings flag

PR #433, opened 2021-03-22
add support for MKX200 projection in DeoVR

crwxaj:
The last update of DeoVR removed the fisheye switch from the option panel, which could somewhat correct the projection of certain SLR Originals releases. This PR sends the correct option to DeoVR for file names containing "mkx200".

### DIFF
--- a/pkg/api/deovr.go
+++ b/pkg/api/deovr.go
@@ -316,6 +316,11 @@ func (i DeoVRResource) getDeoScene(req *restful.Request, resp *restful.Response)
 		})
 	}
 
+	if videoFiles[0].VideoProjection == "mkx200" {
+		stereoMode = "sbs"
+		screenType = "mkx200"
+	}
+
 	if videoFiles[0].VideoProjection == "180_sbs" {
 		stereoMode = "sbs"
 		screenType = "dome"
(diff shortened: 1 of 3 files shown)


theRealKLH:
You should add MKX220 & VRCA220 as well.

crwxaj:
I considered this, but I checked SLR and they only use MKV200.

theRealKLH:
there are at least 2 MKX220's in there. Lesbian Date Parts 1 & 2..

flightsim73:
I’ve updated to the latest version and the screenType is showing MKX200 in the DeoVR JSON but unfortunately it’s still not working. Is there anything else special that you need to do to get these to play properly?

hjs0007:
Some older stasyqvr are also using VRCA220 on SLR.

crwxaj:
> I’ve updated to the latest version and the screenType is showing MKX200 in the DeoVR JSON but unfortunately it’s still not working. Is there anything else special that you need to do to get these to play properly?

Have you restarted DeoVR? It seems to cache some scene data between plays.

crwxaj:
> Some older stasyqvr are also using VRCA220 on SLR.

This got pointed out to me shortly after the release. I'll add MKX220 and VCRA220 for the next version.

flightsim73:
Thanks for following up. It’s working now. I had to delete and reinstall DeoVR on my device.

PR #437, opened 2021-03-25
Enforce yuv420p format to fix error with 7k BadoinkVR videos

lockedmouse:
This PR fixes the preview generation of 7k BadoinkVR videos.

Currently, the 7k BadoinkVR videos are converted into pixel format `yuv420p10le`, while all other videos are converted into pixel format `yuv420p`.

The `yuv420p10le` videos cause the following problems:
- DeoVR immediately freezes when hovering over a BadoinkVR 7k scene. You need to force close it and when restarting, be careful not to hover over a BadoinkVR scene.
- The Firefox video renderer crashes, meaning that all video playback (on all websites) is broken until Firefox is restarted.
- Thumbnails are missing in Windows Explorer: https://i.imgur.com/n1tGRN4.jpg (NSFW)

However, these videos play fine in VLC and Chromium-based browsers.

This PR ensures that all videos are converted into `yuv420p` pixel format.


cld9x:
nice one!

PR #445, opened 2021-03-28
Ignore old search results if they arrive out of order

lockedmouse:
Fixes a bug where the search results are incorrect if the responses arrive out of order. This affects both file matching and quick find.

### DIFF
--- a/ui/src/QuickFind.vue
+++ b/ui/src/QuickFind.vue
@@ -80,6 +80,8 @@ export default {
   data () {
     return {
       data: [],
+      dataNumRequests: 0,
+      dataNumResponses: 0,
       selected: null,
       isFetching: false
     }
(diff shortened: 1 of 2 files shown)


cld9x:
@lockedmouse I am not 100% sure if this is the best way to do it? let me dig into it first before taking it as is

lockedmouse:
Sure. It worked perfectly in my testing and I'm pretty happy with the code quality but I guess there are two ways to improve performance:
1. `dataNumRequests` and `dataNumResponses` are reactive properties because they are declared on the `data` method, so changing them will always trigger a rerender even when it isn't needed.
2. We can store all requests in a list and if a response arrives, abort all older requests. This improves CPU/network speed at the expense of more memory, so I'd say this is an premature optimization.

IMO, the out-of-order search is big enough of a bug that the PR can be merged as-is but if you want to, I can implement 1. I don't see a need to implement 2 at this time.

lockedmouse:
@cld9x Any update on this? I stumble upon this bug almost daily and it is pretty annoying. Right now, when searching I have to type everything except the last character, wait a second and then enter the last character as a workaround.

Please let me know what I can do to get this merged.

PR #446, opened 2021-03-28
Add scripted icon in scene card, fix video previews

lockedmouse:
This PR adds a small icon in the scene card if the scene is scripted. I used Material Design's `pulse` icon which matches the icon we show in the scene details.

Also, this PR changes how video previews work:
- Tags (rating, is_watched, is_scripted, is_multipart) are now displayed on the video preview. They don't take up much space so I don't see why we would want to hide them.
- Fixes a bug where the video previews would keep playing after mouseout. I haven't been able to find the root cause (seems to be related to how the mouse events fire), so this just hides the videos with CSS for now.
- Fixes a bug with some older VirtualRealPorn scenes that are not in 2:1 format (the previews get generated as 400x450, not 400x400). The preview videos are now shrunk to 100% height so that they do not mess up the layout.

### DIFF
--- a/ui/src/views/scenes/SceneCard.vue
+++ b/ui/src/views/scenes/SceneCard.vue
@@ -7,24 +7,22 @@
            @mouseover="preview = true"
            @mouseleave="preview = false">
         <video v-if="preview && item.has_preview" :src="`/api/dms/preview/${item.scene_id}`" autoplay loop></video>
-        <div v-else>
-          <div class="overlay align-bottom-left">
-            <div style="padding: 5px">
-              <b-tag v-if="item.is_watched">
-                <b-icon pack="mdi" icon="eye" size="is-small"/>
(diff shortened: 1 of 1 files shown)


cld9x:
@lockedmouse when preview video is smaller than card area, video doesn't fill up all available space - think this is caused by limiting `max-height` to 100%

lockedmouse:
This already happens in v0.3.8 and is not caused by my PR: https://i.imgur.com/HvIeSxy.jpg (SFWish).

But I agree that this looks wrong, so I'll push a commit shortly to ensure that the video always fills up the available space.

cld9x:
I haven't realized that's a problem until now, so apologies for blaming your commit 🙃

lockedmouse:
Fixed in 8f31aea1ab8274913be4306441831bde74742d05.

PR #448, opened 2021-03-28
Show warning when playing HEVC videos in Firefox

lockedmouse:
Firefox [does not support HEVC videos](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs#hevc-foot-2). Instead, playing a HEVC file only shows a black screen (audio still works).

This PR adds a warning message so that users are not wondering why they cannot see the video.


crwxaj:
HEVC doesn't play in Chrome either.

lockedmouse:
My bad, I only checked Microsoft Edge where it works for me but apparently only because I installed the HEVC extension from the Windows store.

Ideally, we'd do feature detection (`document.createElement('video').canPlayType('video/mp4; codecs="hev1"')`) but Edge just returns an empty string (= cannot play video) when it should say `"probably"`, so let's close this PR for now.

PR #458, opened 2021-04-01
updated go.mod

crwxaj:
Everytime I build these dependencies show up in my go.mod. I avoid committing them, since no one else seems to care, so I thought it might be only my setup. But it's getting annoying. Can we add them to the master branch, or does this cause problems?


crwxaj:
duplicate of #457

PR #465, opened 2021-04-06
show title on scene card

crwxaj:
Some studios don't put the scene title on the cover image. This PR puts the title on the scene card in a unobtrusive way.


cld9x:
Sorry, I don't like it as it adds a lot of noise - I am not against the idea per se, it just require a little bit thoughtful changes in current card layout

Issue #485, opened 2021-04-19
display also Scene ID on scene card at scenes page


Sinoh726:
Team,

can you please add also Scene ID on scenes page (on page where are Tags, Files, Cuepoints, Description displayed).
This information is specially useful for scenes scraped via r18.com

Thank you!

PR #487, opened 2021-04-23
show scene id in scene details

crwxaj:
I agree with #485 that knowing the scene ID is useful sometimes. I tried to make it subtle.


theRealKLH:
fixes #485

PR #488, opened 2021-04-24
add cast list to studio name in DeoVR

crwxaj:
We've got one more line of text in the DeoVR UI. Let's use it.


crwxaj:
This PR doesn't make much sense at the moment since after the layout changes in DeoVR 9.03 the studio line gets cut off in the middle of the available space.

cld9x:
uh, I am going to close it then

PR #489, opened 2021-04-24
Implement cuepoint deletion, add tooltips

lockedmouse:
Adds a button for deleting cuepoints.

delete-cuepoint

Also adds tooltips to various buttons.

Finally, sorts the cuepoints in the DeoVR API. The cuepoints are already sorted chronologically in the XBVR UI, so the user has every reason to believe they'll be sorted the same way in DeoVR.

### DIFF
--- a/ui/src/views/scenes/Details.vue
+++ b/ui/src/views/scenes/Details.vue
@@ -192,7 +192,7 @@
 <script>
 import ky from 'ky'
 import videojs from 'video.js'
-import vr from 'videojs-vr/dist/videojs-vr.min.js'
+import 'videojs-vr/dist/videojs-vr.min.js'
 import { format, formatDistance, parseISO } from 'date-fns'
 import prettyBytes from 'pretty-bytes'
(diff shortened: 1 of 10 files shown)


ui/src/views/scenes/Details.vue
```diff
@@ -301,7 +304,7 @@ export default {
     updatePlayer (src, projection) {
       this.player.reset()
 
-      const vr = this.player.vr({
+      /* const vr = */ this.player.vr({
```
cld9x, line 307: @lockedmouse why's that?

lockedmouse: ESLint shows a warning that `vr` is an unused variable. I didn't want to remove it totally because six lines below it is still used in commented-out code:

```js
// vr.camera.position.set(-1, 0, 2);
```

IMO this is the best compromise that removes the ESLint warning but preserves the code in case someone wants to uncomment the camera position. Personally, I am a proponent of deleting code over commenting it out, so alternatively I can just remove both comments if you want.

PR #491, opened 2021-04-24
Implement "skip intro" feature

lockedmouse:
With the `skipIntro` property from the DeoVR API, we can skip the studio-specific intro. Users need to enable "skip intro" in their DeoVR settings for this to work. Existing scenes will need to be rescraped in order for the intro length to be set correctly.

Also fixes a bug where when adding a cuepoint, the new cuepoint shows up in the scene popup, but when closing and reopening the scene popup, the cuepoint is gone until the page is reloaded.

This PR is WIP until I get around to adding the intro durations for more studios.


lockedmouse:
As an update, when testing this PR I ran into problems where DeoVR would refuse to playback videos with the `skipIntro` property. Like it would load forever and show a frozen image.

Not yet sure whether this happened due to the PR or my testing environment or DeoVR is just bugged. Will pick this up again sometime and hopefully figure out what went wrong.

PR #506, opened 2021-05-09
prevent crash in VRBangers scraper

crwxaj:
For some reason the scraper request to https://vrbangers.com/video/final-sexam/ gets the content for https://vrbangers.com/japanese/. I couldn't figure out why and couldn't reproduce it in the brower. This PR at least prevents the resulting crash.

### DIFF
--- a/pkg/scrape/vrbangers.go
+++ b/pkg/scrape/vrbangers.go
@@ -29,6 +29,7 @@ func VRBangersSite(wg *sync.WaitGroup, updateSite bool, knownScenes []string, ou
 		sc.HomepageURL = strings.Split(e.Request.URL.String(), "?")[0]
 
 		if e.ChildAttr(`link[rel=shortlink]`, "href") == "" {
+			log.Printf("Skipping %s because it is not a valid scene.", e.Request.URL.String())
 			return
 		}
 		// Scene ID - get from URL
--- a/pkg/scrape/vrbangers.go
+++ b/pkg/scrape/vrbangers.go
@@ -28,6 +28,9 @@ func VRBangersSite(wg *sync.WaitGroup, updateSite bool, knownScenes []string, ou
 		sc.Site = siteID
 		sc.HomepageURL = strings.Split(e.Request.URL.String(), "?")[0]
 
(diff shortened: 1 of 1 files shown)


WilsonBradley:
Sorry to be a noob, how to apply? The xbvr.exe
 crashes right away.

pkg/scrape/vrbangers.go
```diff
@@ -28,6 +28,9 @@ func VRBangersSite(wg *sync.WaitGroup, updateSite bool, knownScenes []string, ou
 		sc.Site = siteID
 		sc.HomepageURL = strings.Split(e.Request.URL.String(), "?")[0]
 
+		if e.ChildAttr(`link[rel=shortlink]`, "href") == "" {
+			return
```
lockedmouse, line 32: Looks like this was a temporary glitch as the VRBangers scraper no longer crashes.

Maybe add some logging so that we can at least see that the scene is skipped?

```suggestion
			log.Printf("Skipping %s because it is not a valid scene.", e.Request.URL.String())
			return
```

lockedmouse:
So it looks like what happened is that VRBangers has secretly added 14 new uncensored JAV scenes filmed between December 2017 and June 2018. They've always been available on SLR (e.g. [here](https://www.sexlikereal.com/scenes/train-ticket-to-heaven-4472)) but only last week they got added to VRBangers.com as well.

On the first day, all video URLs (e.g. [this one](https://vrbangers.com/video/final-sexam/)) automatically redirected to [`/japanese/`](https://vrbangers.com/japanese/) which is why our scraper was having problems. They've since fixed it and those scenes now show up correctly.

There still is a problem with the model pages, e.g. for Suzumiya Kotone the scene page links to [`/?post_type=model&p=11616`](https://vrbangers.com/?post_type=model&p=11616) which is a 404, but when you go to [`/model/suzumiya-kotone/`](https://vrbangers.com/model/suzumiya-kotone/) (which should be the correct URL) it redirects to `/japanese/`.

What I don't understand is why VRBangers always keeps quiet about this stuff. There's nothing mentioned on their forums or any other places I can see. You'd expect that when adding new scenes for their members, they'd want to shout it from the rooftops...

PR #509, opened 2021-05-10
Added SLR scraper for VRLatina/HoloGirlsVR/VRPFilms/RealityLovers

CoolMintChocolate:
Adding SLR as an alternative source for VRLatina scenes, because there are scenes that are only on SLR and not on VRLatina website. If the user chooses to use both the SLR scraper and the dedicated VRLatina scraper, the scenes that come from SLR will have `vrlatina-slr` prefix in its id, while the scenes from the dedicated scraper will have its existing `vrlatina` prefix.

I had to change the SLR scraper a bit to make this prefix happen, in the current code base it's a convention that SLR scraper always uses the scraper Id in the scene Id, which will lead to `vrlatina` being used for both SLR & dedicated scraper, which might lead to ID duplications.

Edited: I've added support for HoloGirlsVR/VRPFilms/RealityLovers together with this PR, because they fundamentally do the same thing code-wise.

### DIFF
--- a/pkg/scrape/slrstudios.go
+++ b/pkg/scrape/slrstudios.go
@@ -14,7 +14,7 @@ import (
 	"github.com/xbapps/xbvr/pkg/models"
 )
 
-func SexLikeReal(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out chan<- models.ScrapedScene, scraperID string, siteID string, company string) error {
+func SexLikeReal(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out chan<- models.ScrapedScene, slrId string, scraperID string, siteID string, company string) error {
 	defer wg.Done()
 	logScrapeStart(scraperID, siteID)
 
@@ -219,7 +219,7 @@ func SexLikeReal(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out 
 		}
 	})
 
-	siteCollector.Visit("https://www.sexlikereal.com/studios/" + scraperID + "?sort=most_recent")
+	siteCollector.Visit("https://www.sexlikereal.com/studios/" + slrId + "?sort=most_recent")
 
 	if updateSite {
 		updateSiteLastUpdate(scraperID)
@@ -229,6 +229,10 @@ func SexLikeReal(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out 
 }
 
 func addSLRScraper(id string, name string, company string, avatarURL string) {
+	addSLRScraperWithAlternativeId(id, id, name, company, avatarURL)
+}
+
+func addSLRScraperWithAlternativeId(slrId string, id string, name string, company string, avatarURL string) {
 	suffixedName := name
 	if company != "SexLikeReal" {
 		suffixedName += " (SLR)"
@@ -239,7 +243,7 @@ func addSLRScraper(id string, name string, company string, avatarURL string) {
 	}
 
 	registerScraper(id, suffixedName, avatarURL, func(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out chan<- models.ScrapedScene) error {
-		return SexLikeReal(wg, updateSite, knownScenes, out, id, name, company)
+		return SexLikeReal(wg, updateSite, knownScenes, out, slrId, id, suffixedName, company)
 	})
 }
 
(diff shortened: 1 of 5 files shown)


lockedmouse:
How does the second scraper show up in XBVR? Ideally, the scraper table should contain two entries, "VRLatinas" and "VRLatinas (SLR)", both with a separate toggle and the ability to force update and delete.

Unfortunately, right now the force update/delete is buggy (fixed in #351 and #479) so please make sure this PR is compatible to those two PRs.

pkg/scrape/slrstudios.go
```diff
@@ -176,12 +176,16 @@ func SexLikeReal(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out
 }
 
 func addSLRScraper(id string, name string, company string, avatarURL string) {
+	addSLRScraperWithAltenativeId(id, id, name, company, avatarURL)
+}
+
+func addSLRScraperWithAltenativeId(slrId string, id string, name string, company string, avatarURL string) {
```
crwxaj, line 182:
```suggestion
func addSLRScraperWithAlternativeId(slrId string, id string, name string, company string, avatarURL string) {
```

pkg/scrape/slrstudios.go
```diff
@@ -176,12 +176,16 @@ func SexLikeReal(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out
 }
 
 func addSLRScraper(id string, name string, company string, avatarURL string) {
+	addSLRScraperWithAltenativeId(id, id, name, company, avatarURL)
```
crwxaj, line 179:
```suggestion
	addSLRScraperWithAlternativeId(id, id, name, company, avatarURL)
```

pkg/scrape/slrstudios.go
```diff
@@ -228,6 +232,7 @@ func init() {
 	addSLRScraper("vredging", "VRedging", "VRedging", "https://mcdn.vrporn.com/files/20200630081500/VRedging_LOGO_v1-400x400.jpg")
 	addSLRScraper("vrextasy", "VReXtasy", "VReXtasy", "https://www.sexlikereal.com/s/refactor/images/favicons/android-icon-192x192.png")
 	addSLRScraper("vrfirsttimer", "VRFirstTimer", "VRFirstTimer", "https://mcdn.vrporn.com/files/20200511115233/VRFirstTimers_Logo.jpg")
+	addSLRScraperWithAltenativeId("vrlatina", "vrlatina-slr", "VRLatina", "VRLatina", "https://pbs.twimg.com/profile_images/979329978750898176/074YPl3H_200x200.jpg")
```
crwxaj, line 235:
```suggestion
	addSLRScraperWithAlternativeId("vrlatina", "vrlatina-slr", "VRLatina", "VRLatina", "https://pbs.twimg.com/profile_images/979329978750898176/074YPl3H_200x200.jpg")
```

crwxaj (commented):
Small typo fix

lockedmouse:
Oh, then my concern was unwarranted; this looks fine. 👍

I'm a little concerned that the scenes are prefixed by `vrlatinas-` and `vrlatinas-slr-`, so it will be difficult to query for scenes of the dedicated VRLatinas scraper without also getting SLR scenes. Maybe change it to `vrlatinas-main-` or similar?

CoolMintChocolate:
I wouldn't be opposed to that at all - the only possible concern I have is that we might have to migrate existing data to new studio id, right? I'm not overly familiar with the database structure at the moment, so that's just a guess. If we do have to migrate data, is there some code in the current code base that did something similar before so I can repurpose it for this PR?

CoolMintChocolate:
@lockedmouse @crwxaj I have added migration path for existing scenes scraped by the existing VRLatina scraper. From now on scenes scraped by that scraper will have `vrlatina-site-` prefix, while scenes scraped by SLR scraper will have `vrlatina-slr-` prefix. I have also tested that deleting scenes scraped by one scraper does not affect scenes scraped by the other scraper.

CoolMintChocolate:
I've also added HoloGirlsVR/VRPFilms/RealityLovers SLR scrapers to this PR as well, they fundamentally do the same thing as the VRLatina SLR scraper

alexnsfw:
In the migrations, actions also need to be updated to refer to scenes with their new IDs. Also scene's ID changes it needs to be re-indexed, so you could kick off the indexing at the end of the migration. See #586 for an example.

CoolMintChocolate:
Thanks for pointing that out @alexnsfw. I've made the changes necessary to address your concerns.

CoolMintChocolate:
Closing this PR since this feature is now available out of the box

PR #511, opened 2021-05-11
Update github workflows to build docker containers and build arm7, arm8 and x86_64 containers.

### LINKED ISSUE #252: Arm based Docker builds
I have an arm based nas (helios 4) and need a arm docker image.
To get this to work I needed to make 2 changes:
Line 8 you are downloading go with curl but the linux-amd64 arch.
Could we add an ARCH variable as I need to download linux-armv6l when building the image.

The other problem I have is you are using distoless base image and there is no arm image.
Changing this to a basic debian image allows me to build and run this inside a container.
I can see the reason for using this but is there a similar project that supports arm as well as amd64?

Tweeticoats:
This updates the github actions and creates a docker build action.
This uses buildx to build the container for arm7, arm 8 and x86_64 in parallel and pushes this to the docker hub.
 This should resolve #252

To use this you need to configure secrets in github to pass the username and password to login to the docker hub.
To configure this under the github repository settings > Secrets > Action Secrets, and create 2 new secrets.
DOCKERHUB_USERNAME
DOCKERHUB_TOKEN

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,3 +27,26 @@ jobs:
       
     - name: Check code formatting using gofmt
       uses: Jerome1337/gofmt-action@v1.0.4
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Checkout
+      run: git fetch --prune --unshallow --tags
+    - name: Login to Docker Hub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+    - name: Development Docker
+      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+      env:
+        DOCKER_CLI_EXPERIMENTAL: enabled
+      run: |
+        docker run --privileged --rm tonistiigi/binfmt --install all
+        docker info
+        docker buildx create --name builder --use
+        docker buildx inspect --bootstrap
+        docker buildx ls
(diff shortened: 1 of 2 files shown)


cld9x:
@Tweeticoats looks sweet, I already made some first steps towards moving to Github Actions so I'll need to review and merge these two efforts together

atani12:
Thanks @Tweeticoats for the code.

I just changed my setup to ARM and realised the docker image was only for amd64 :(
Fortunately I took your commit to make a customised one, and not knowing what was Github actions until today some searching on the web helped me adapt the code and get a working docker image for arm running :D

I just got it working but noticed that ffprobe will not work, giving the error `time="2021-07-11T14:05:56Z" level=error msg="Error running ffprobe/...fork/exec /root/.config/xbvr/bin/ffprobe: exec format error`. I guess it's expecting a amd64 instruction?
Just wanted to share that in case it's not been spotted yet on a ARM docker image.

Cheers

Tweeticoats:
@atani12  Good point, this does not fix downloading ffprobe and ffmpeg for x86_64 and that does not work on arm.
The workaround is to download static ffmpeg binaries from https://johnvansickle.com/ffmpeg/ and place this in the bin folder.
This is what xbvr does first startup and it is something that I did manually the first time and forgot about it.
So there will need to be a change to xbvr to see what architecture it is running under and download the right ffmpeg.

docker cp ffprobe xbvr:/root/.config/xbvr/bin/
docker cp ffmpeg xbvr:/root/.config/xbvr/bin/

atani12:
Thanks for the workaround :)

theRealKLH:
The code is in there to pull the correct binaries for ARM. Maybe you pointed ARM docker to your already existing AMD64 folders?

Tweeticoats:
pull request no longer useful

PR #517, opened 2021-05-20
fix "delete scraped scenes / force update scenes"

theRealKLH:
fixes "delete scenes" & "force update scenes" on SLR, VRPorn, PornVR based scrapers


theRealKLH:
Previously (and better solution) submitted #351

PR #529, opened 2021-06-02
Fix POVR titles containing a K

alexnsfw:
This fixes a bug where titles of POVR scenes would get cut off if they contain a 'K' (e.g. 'Fountain of Knowledge' would become 'Fountain of ').


theRealKLH:
good catch.

PR #533, opened 2021-06-03
Fix CzechVR scraper adding 'Interactive' to cast

alexnsfw:
Since CzechVR added scripts, the scraper has been adding 'Interactive' to the cast of a bunch of scenes.

### DIFF
--- a/pkg/scrape/czechvr.go
+++ b/pkg/scrape/czechvr.go
@@ -64,7 +64,7 @@ func CzechVR(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out chan
 		})
 
 		// Cast
-		e.ForEach(`div.post div.featuring a`, func(id int, e *colly.HTMLElement) {
+		e.ForEach(`div.post div.featuring > a`, func(id int, e *colly.HTMLElement) {
 			sc.Cast = append(sc.Cast, strings.TrimSpace(e.Text))
 		})
 


theRealKLH:
you got an example of this happening?

alexnsfw:
Czech VR 415, for example. The HTML looks like this:
```html
<div class="featuring">
    Featuring
    <a href="./model-anna-de-ville">Anna de Ville</a>	
    <div class="interactive">
        <a href="./tag-teledildonics" title="Interactive toys">Interactive</a>
    </div>
</div>
```
So the current selector incorrectly selects the second link:
Screenshot_2021-06-03 XBVR

theRealKLH:
that's not happening on mine. 
`<div class="interactive> <a href="huh" title="interactive">Interactive</a>` 

isn't part of 
`<div class="featuring"> <a href="blahblahblah">actress a</a> <a href="yeahyeahyeah">actress b</a>`

image

Only place I'm seeing "interactive" is tags on WankzVR & MilfVR

alexnsfw:
Was that scraped recently? Scripts were only added in the last week or so, to older scenes, so if it was scraped when the scene released in April I would expect it to be fine. And none of the scenes released since then have scripts so they don't exhibit the issue.

I see `div.interactive` inside of `div.featuring` no matter how I fetch the page (cache or not in firefox, chrome, curl, this scraper...)

theRealKLH:
That scene would have been scraped right around the time it was released, so no later than 4/24 or 4/25. I am going to start with a clean DB and try...

I'm not a coding expert. Still learning for this. It definitely shows at times and I can flub with the best of em, But the way I look at the code is "yes" `div.interactive` is inside of `div.featuring`, but it's not inside `div.featuring a`, so it shouldn't be picked up unless targeting `div.featuring div.interactive a`, which we're not.. "no"? `div.featuring a` can't just skip over the entire `<a href  /a> <div.interactive>` can it? 

And I just finished the new scrape and YUP... it is picking it up as you say. Thanks for the lesson, this may help me somewhere else...

alexnsfw:
Ah, yes, that's the difference that `>` makes.

The space between `div.featuring` and `a` is the [descendant combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator), so all `a` elements anywhere inside of `div.featuring` are selected. `>` is the [child combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator), so `div.featuring > a` only selects `a` elements which have `div.featuring` as their (immediate) parent.

PR #539, opened 2021-06-06
Add wishlist

alexnsfw:
This PR adds a wishlist. It's just like the watchlist or favourites, except:
* available scenes cannot be added to it; and
* when a scene becomes available, it's removed from the wishlist.

### DIFF
--- a/pkg/api/options.go
+++ b/pkg/api/options.go
@@ -46,6 +46,7 @@ type RequestSaveOptionsWeb struct {
 	TagSort           string `json:"tagSort"`
 	SceneWatchlist    bool   `json:"sceneWatchlist"`
 	SceneFavourite    bool   `json:"sceneFavourite"`
+	SceneWishlist     bool   `json:"sceneWishlist"`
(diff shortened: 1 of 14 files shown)


maxcom99:
Can this please be merged in?

alexnsfw:
> Can this please be merged in?

I'm still holding out hope 2 years later...

Maintainers, is there a chance?
(@crwxaj maybe?)

crwxaj:
To be honest I haven't really considered it. This PR was from before I was at the helm.
I'm not really against it, but couldn't this be easily emulated with a "Watchlist" and "Not downloaded" filter?

alexnsfw:
> couldn't this be easily emulated with a "Watchlist" and "Not downloaded" filter?

That'll get you pretty close (I think that's what I used to do before implementing this), but it's not quite the same. Personally, I add scenes to my wishlist that I don't necessarily want to be in my watchlist when I do download them. Plus I like having the different color toggle button and the one-click wishlist filter.

PR #542, opened 2021-06-08
Fix VRPorn cast scraping

alexnsfw:


theRealKLH:
looks like I forgot to submit my PR for this (https://github.com/theRealKLH/xbvr/tree/vrporn_062021). Take a look. There is also a change needed for retrieving the scene date from "premium" pages. I also fixed the code for "duration", however a lot of the pages only provide "trailer" length not full video time.

alexnsfw:
Oh I hadn't noticed the date issue, good catch. :+1:

As for the duration, I noticed that it's broken for a bunch of scrapers, so I've been working on fixing as many as I can and I plan on grouping them all in one PR. For this one I thought about just removing the duration when only the trailer's length is available, since that doesn't seem very useful (especially if you want to filter on the duration, which is what I want to implement after fixing all those scrapers).

Any particular reason why you changed from parsing the duration from the HTML to parsing it from the script? Is it because it's the same regardless of if the full length is available or only that of the trailer?

Anyway, if you want to submit your changes as a PR, go ahead and I'll close this one. Or I can include the date fix in this PR and the duration fix in my upcoming PR. I don't really mind either way. :slightly_smiling_face:

theRealKLH:
> Oh I hadn't noticed the date issue, good catch. 👍
> 
> As for the duration, I noticed that it's broken for a bunch of scrapers, so I've been working on fixing as many as I can and I plan on grouping them all in one PR. For this one I thought about just removing the duration when only the trailer's length is available, since that doesn't seem very useful (especially if you want to filter on the duration, which is what I want to implement after fixing all those scrapers).
> 
> Any particular reason why you changed from parsing the duration from the HTML to parsing it from the script? Is it because it's the same regardless of if the full length is available or only that of the trailer?
> 
> Anyway, if you want to submit your changes as a PR, go ahead and I'll close this one. Or I can include the date fix in this PR and the duration fix in my upcoming PR. I don't really mind either way. 🙂

Just realized I never answered you. When I began looking at the scraper, ".lengthvideoAfter" & ".lengthvideoBefore" didn't exist (typo?). Even after changing them to ".lenghvideoAfter" & ".lenghvideoBefore" I kept coming up with duration = 0 after scraping. Appears to be creating final html "on the fly" from script, so I changed to "parsing duration from script".

```e.DOM.Find(`.lenghvideoAfter.premium-post`)```  There is no ```<div class = "lenghvideoAfter premium-post" blah blah blah </div>``` in the html.

As I look back at this, I guess any scene that's not "premium" only has a trailer.

alexnsfw:
So... I incorporated more fixes into this PR: cast, release date and gallery. Scraping was also completely broken for some studios (e.g. Tonight's Girlfriend), so I fixed that as well. I'll include the duration fix in a separate PR.

PR #562, opened 2021-08-18
vrbangers - don't pull paypal images

theRealKLH:
Paypal image was getting scraped as a "cover"

### DIFF
--- a/pkg/scrape/vrbangers.go
+++ b/pkg/scrape/vrbangers.go
@@ -92,7 +92,7 @@ func VRBangersSite(wg *sync.WaitGroup, updateSite bool, knownScenes []string, ou
 			sc.Covers = append(sc.Covers, e.ChildAttrs(`section.banner picture img`, "data-pagespeed-lazy-src")[0])
 		}
 
-		if len(e.ChildAttrs(`section.base-content--bg img`, "data-pagespeed-lazy-src")) > 0 {
+		if len(e.ChildAttrs(`section.base-content--bg div.base-border img`, "data-pagespeed-lazy-src")) > 0 {
 			sc.Covers = append(sc.Covers, e.ChildAttrs(`section.base-content--bg img`, "data-pagespeed-lazy-src")[0])
 		}
 


pkg/scrape/vrbangers.go
```diff
@@ -92,7 +92,7 @@ func VRBangersSite(wg *sync.WaitGroup, updateSite bool, knownScenes []string, ou
 			sc.Covers = append(sc.Covers, e.ChildAttrs(`section.banner picture img`, "data-pagespeed-lazy-src")[0])
 		}
 
-		if len(e.ChildAttrs(`section.base-content--bg img`, "data-pagespeed-lazy-src")) > 0 {
+		if len(e.ChildAttrs(`section.base-content--bg div.base-border img`, "data-pagespeed-lazy-src")) > 0 {
 			sc.Covers = append(sc.Covers, e.ChildAttrs(`section.base-content--bg img`, "data-pagespeed-lazy-src")[0])
 		}
```
alexnsfw, line 97: The same selector should probably be used when retrieving the cover:
```suggestion
		if len(e.ChildAttrs(`section.base-content--bg div.base-border img`, "data-pagespeed-lazy-src")) > 0 {
			sc.Covers = append(sc.Covers, e.ChildAttrs(`section.base-content--bg div.base-border img`, "data-pagespeed-lazy-src")[0])
		}
```
Or better yet avoid repeating it with something like this:
```suggestion
		if tmpCovers := e.ChildAttrs(`section.base-content--bg div.base-border img`, "data-pagespeed-lazy-src"); len(tmpCovers) > 0 {
			sc.Covers = append(sc.Covers, tmpCovers[0])
		}
```

theRealKLH:
Good call.. in my "rush" to fix the mistake I didn't "fix the mistake" Grin. I will fix once I'm back at a computer..

Just wondering though... Wouldn't the test for "div.base-order" restrict the append to JUST that current img?

alexnsfw:
> Wouldn't the test for "div.base-order" restrict the append to JUST that current img?

The `if` checks that there's at least one image in `div.base-border` and if there is it appends the first image in all of `section.base-content--bg`. It works (I think) because the image in `div.base-border` (the cover) is the first one in that section, but it's just not obvious.

theRealKLH:
Thanks, your suggestions actually uncovered some inconsistency in the web code, so I made additional change..

pkg/scrape/vrbangers.go
theRealKLH: Thanks, your suggestions actually uncovered some inconsistency in the web code, so I made additional change..

theRealKLH:
https://github.com/xbapps/xbvr/issues/142#issuecomment-902030312

PR #568, opened 2021-08-28
fix scraper actions for SLR/VRPorn sites

crwxaj:


theRealKLH:
heh... dupe of #351?

crwxaj:
You are right. I didn't check that. I will leave it open though, since #351 has conflicts now and I don't think it will get merged anytime soon.

PR #619, opened 2021-11-27
Use actual slug for SLR not the internal ID

SorenSkovSkov:
In order to have competing scrapers for sites like VRPFilms and VRLatina where the site and the SLR catalogue don't always have the same videos, it's necessary to change the internal ID for the site. This PR adds that option without affecting the slug used for scraping.


theRealKLH:
example? of how this would look/work..? I'm being dense trying to visualize.

SorenSkovSkov:
In my local branch I have the following two SLR sites:
```
addSLRScraper("vrlatina-slr", "vrlatina", "VRLatina", "VRLatina", "https://cdn-vr.sexlikereal.com/images/studio_creatives/logotypes/1/110/logo_crop_1606820822.png")
addSLRScraper("vrpfilms-slr", "vrpfilms", "VRPFilms", "VRPFilms", "https://vrpfilms.com/storage/settings/March2021/Z0krYIQBMwSJ4R1eCnv1.png")
```
As you can see, can give their IDs the `-slr` suffix, so they don't interfere with the existing scrapers for those two sites, without it being used to scrape `"https://www.sexlikereal.com/studios/" + scraperID + "?sort=most_recent"` like it would in the master branch.

theRealKLH:
OK (I see it now). looked more closely. It seems this is the same as what's being done (for the most part) in PR #509 (which I hadn't looked at until now). That code only needs "slug" entry to the "non" SLR sites (via AddAlternativeScraper function). Although the true site scrapers are also being updated. Not sure if that is necessary (didn't look that close). 

How come you "return" slug twice here instead of id & slug? 

`return SexLikeReal(wg, updateSite, knownScenes, out, slug, name, company, slug)`

doesn't that wind up ignoring scraperID (vrlatina-slr)? causing conflict w/main site since it is only using slug (vrlatina)?

SorenSkovSkov:
Yes, it's the same principle as #509, just with less changes.
The double `slug` was a mistake, thanks for catching that.

theRealKLH:
I would suggest more like that PR, only exceptions would require the additional "slug" element, so as to not have to update every single studio (those ONLY hosted on SLR).

PR #622, opened 2021-12-05
Add funcsript speed features

agentfisk:
Calculates median speed of each script file and saves it on the file model. 
Finds selected script on scene and saves speed on the model. 

Adds funscript speed desc/asc and range filters. 
Screenshot from 2021-12-05 14-45-10
Screenshot from 2021-12-05 14-44-57

Displays scenes speed on the card.
Screenshot from 2021-12-05 14-46-47
Not sure about the look of this one.

Adds option to prepend speed to title in DeoVR API, at the loss of matching local exported scripts.
Screenshot from 2021-12-05 14-44-25

### DIFF
--- a/pkg/api/deovr.go
+++ b/pkg/api/deovr.go
@@ -385,20 +385,24 @@ func (i DeoVRResource) getDeoScene(req *restful.Request, resp *restful.Response)
 		screenType = "sphere"
 	}
 
-	// title := scene.Title
+	title := scene.Title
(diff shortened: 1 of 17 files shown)


agentfisk:
It's units/sec stroke speed. There is a 100 units between top and bottom position, so i guess a speed of 100 would be half a stroke per sec? Same measurement used by OpenFunscripter and Funscript.io.

I use it to categorize scenes by how fast their scripts. In the case of xbvr i use it to make randomized DeoVR playlists within certain speed thresholds, using the funscript min/max filters, and one playlist that just sort by slowest scenes.

PR #633, opened 2021-12-29
Fix GitPod environment setup

jwvanderbeck:
I am NOT a GitPod/Go developer so comments/suggestions/requested changes are welcome.

This PR updates the Gitpod.yml file to run a series of commands before `yarn dev` to get the environment in the proper state for `yarn dev` to run successfully.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -4,12 +4,6 @@ tasks:
   - name: Continuous Build
     command: |
       cd /workspace/xbvr
-      yarn
-      go mod download github.com/ProtonMail/go-appdir
-      go get
-      go generate
-      go get
-      yarn dev
 ports:
   - port: 9999
     onOpen: open-preview
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -5,7 +5,6 @@ tasks:
     command: |
       cd /workspace/xbvr
       yarn install && yarn build && go mod download github.com/ProtonMail/go-appdir && go generate && go get
-      yarn dev
 ports:
   - port: 9999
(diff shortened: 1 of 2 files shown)


jwvanderbeck:
I am concerned I may have a bug still because while this appears to set everything up properly and the interface does appear, any subsequent code changes don't necessarily get reflected in the running instance as expected. Sometimes they do, sometimes they don't.  

It looks as though the hot recompile is sometimes not able to restart the server?

```
[dev:go] time="2021-12-29T17:37:37Z" level=info msg="listen tcp 0.0.0.0:9998: bind: address already in use"
[dev:go] time="2021-12-29T17:37:37Z" level=fatal msg="listen tcp 0.0.0.0:9998: bind: address already in use"
```

Sometimes it gives those errors but keeps trying and the changes take affect, other times it just stops trying.  Or at least that is what it seems like in the console.

jwvanderbeck:
After discussing this in Discord with @CoolMintChocolate it seems that it is normal for it to get in this stuck state and need to be manually restarted.  However when the environment is run using `yarn dev` and the two processes handled by `concurrently` the user is unable to ctrl-c and restart the server.

Skipping `yarn dev` and instead starting up `yarn dev:ui` and `yarn dev:go` in separate terminal tabs in GitPod allows the user to restart manually then this stuck state is achieved.

I would recommend to remove the auto start of `yarn dev` from the environment, and update the instructions slightly to indicate to run the two commands manually.

jwvanderbeck:
Is this still relevant after the build system changes or should I just close it?

PR #669, opened 2022-01-15
Fix SexBabesVR CSS

jwvanderbeck:
Small change to the CSS


theRealKLH (commented):
may not make a difference here, but you may want to change line 111 to 

`	siteCollector.OnHTML(`div.c-pagination li.next a`, func(e *colly.HTMLElement) {`

jwvanderbeck:
Yeah I can test that.  I just made the smallest change possible to what was previously there.  I agree with you though so let me test that locally real quick.

Issue #442, opened 2021-03-27
Manual entry?


jwvanderbeck:
I have a scene I downloaded from SLR that I can't seem to get scraped anywhere in the default scrapers, nor by pointing it directly to the scene URL.

Is there some way I can just manually enter this scene so that I can actually watch it?

theRealKLH:
have you tried to manually match it from the files screen?

jwvanderbeck:
Match it to what? There is nothing to match it to because the proper scene can't be scraped through any scrapers that I can find.

theRealKLH:
Is it a scene that should be scrapable? what studio? unless you want to give us the name...

jwvanderbeck:
Hence the desire to be able to manually make an entry without scraping :)

theRealKLH:
currently you'd have to "false match" with an existing scraped scene and edit all the scene data manually.

str0keR:
+1 being able to manually create a scene. I have, for example, seens from SinsVR that are no longer shown on their site so unable to be scraped. I'd like to be able to at least manually match the scene; provide some tags, scene name, cover image etc so I can watch the scene in DeoVR

Noob34567:
How is the progress on this? Create manual entries would be awesome and a good way to handle scrappers that don't work.

theRealKLH:
@jwvanderbeck cleaning "dead" issues. This should be resolved.

PR #683, opened 2022-01-23
Manual scene creation for XBVR

NodudeWasTaken:
This is an attempt to address issue #442 regarding adding custom scene entries.

The changes i made was to add a /create endpoint to the scenes api.
The way i did this was to create an imaginary vr site called "CustomVR".
From there it creates and id based on the current time and adds the scene to the db.
It then proceeds to call scene create and search index update.
It only takes a scene title, as i assume the user can edit the rest into the scene.

I also added a ui entry in the "Scrapers" options section for adding said custom scene. This could potentially be moved to scrapers, or elsewhere.

Some potential and actual problems:

- Duration is unchangeable in the EditPanel UI? Im unsure how important it is.
- Its currently placed in the scrapers section instead of a more convenient place like in the matching ui. This is mainly to discourage its usage, as i assume we'd rather have the scraper actually find the scenes instead of users creating new ones.
- Having a fixed "CustomVR" assumes nobody ever creates an actual "CustomVR" studio.
- Unrelated: The first added scene doesn't get added to the search index? I dont know what thats about, but it needs to be solved.
- Fixed: The Edit panel breaks if there are no cover urls, so in my infinite wisdom i added a fake one, but that also needs to be fixed.
- Fixed: Custom ID generation

Any complaints or improvements are appreciated, i've never written in GO before this.

### DIFF
--- a/pkg/api/scenes.go
+++ b/pkg/api/scenes.go
@@ -134,7 +134,7 @@ func (i SceneResource) createCustomScene(req *restful.Request, resp *restful.Res
 	scene_id := req.QueryParameter("scene_id")
 	if (scene_id == "") {
 		log.Infof("SceneID missing from request!")
-		CustomID = "Custom-" + time.Now().Format("20060102150405")
+		CustomID = "Custom-" + time.Now().Format("2006010215040506")
 	} else {
 		CustomID = scene_id
 	}
--- a/pkg/api/scenes.go
+++ b/pkg/api/scenes.go
@@ -132,9 +132,10 @@ func (i SceneResource) createCustomScene(req *restful.Request, resp *restful.Res
 
 	var CustomID string
 	scene_id := req.QueryParameter("scene_id")
+	currentTime := time.Now()
 	if (scene_id == "") {
 		log.Infof("SceneID missing from request!")
-		CustomID = "Custom-" + time.Now().Format("2006010215040506")
+		CustomID = "Custom-" + currentTime.Format("2006010215040506")
 	} else {
 		CustomID = scene_id
 	}
@@ -147,6 +148,7 @@ func (i SceneResource) createCustomScene(req *restful.Request, resp *restful.Res
 	scene.Site = "CustomVR"
 	scene.HomepageURL = "http://localhost/" + scene.SceneID
 	scene.Covers = append(scene.Covers, "http://localhost/dont_cause_errors")
+	scene.Released = currentTime.Format("2006-01-02")
 
 	log.Infof("Creating custom scene: %v %v", scene.SceneID, scene.Title)
 
--- a/pkg/api/scenes.go
+++ b/pkg/api/scenes.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"strconv"
 	"time"
+	"strings"
 
 	"github.com/go-test/deep"
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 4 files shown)


pkg/api/scenes.go
```diff
@@ -117,6 +121,50 @@ func (i SceneResource) WebService() *restful.WebService {
...
+	title := req.QueryParameter("title")
+	if (title == "") {
+		log.Error("Title is missing from request!")
+		return
+	}
+
+	var unusedCustomID string
+
+	var lastScene models.Scene
+	err := db.Where("site = ?", "CustomVR").Order("SceneID").Last(&lastScene)
```
alexnsfw, line 137: A few issues with this:
1. The column is named `scene_id`, doesn't this just fail (i.e. return on line 147)?
2. The order is implicitly ascending and `Last` is defined as taking the last record ordered by primary key. This result in a query like `... order by scene_id asc, id desc limit 1`... which selects the first custom scene (`Custom-0`) rather than the last one.
3. Ordering by scene ID to get the last custom scene will stop working at the 11th custom scene because `Custom-9 > Custom-10`.
4. Getting the last scene ID and incrementing it by 1 is going to be problematic if the option to delete custom scenes is added (or if one deletes the scene manually from the database). Say I create `Custom-0`, edit it to add a title and description and then delete it. Then I create a new custom scene which gets `Custom-0` again as its scene ID. Now say I edit this one but only set a title, no description. Next time edits get re-applied after scraping, the new `Custom-0` scene will get the description from the one that was deleted.

(I haven't tested any of this in XBVR, but I *have* tested it in a small standalone test program.)

NodudeWasTaken: I did do simple tests that this did work with up to 8 scenes, but you are right that sql will probably do string sorting and it will break (remember to backup the database if doing this).
I think the solution to this is to have an integer only this code reads and writes to, so it always gets incremented, even if all other scenes gets deleted.
Im unsure where to store said integer though, as i see it, there are 2 options.
I make a database table with only 1 entry, where said entry is said iterator.
Or, i make a file in the config folder where i store said integer and write to it.

The advantage of the single file is that it seems wasteful to create a db table for such a simple value.
But the advantage of a db entry, is that its more stable and users are less likely to mess with it.

alexnsfw: Alright, I tested it within XBVR and it does work, even past 11 custom scenes. However it works kind of by accident. `Order("SceneID")` is invalid so GORM *v1* just silently ignores it, so it ends up taking the last scene ordered by `ID` which works just fine since the ID is auto-incremented. GORM *v2*, which I was initially testing with, errors out because of the invalid column name. This accounts for my points 1 to 3.

As for point 4, another option would be to derive the scene ID from the current date and time. Or creating a random one and checking that it doesn't already exist in either the `scenes` table or the `actions` table.

NodudeWasTaken: Alright, so i now made the scene id based on the current time, the norm seemed to be to have a number as id, so i made it into a number (currently its time down to 2 decimals of the millisecond, to attempt to prevent collisions).
Its also accepts a custom id from the user, although this isn't present in the UI.

NodudeWasTaken: A assume that thumbs up means approved 😀

jwvanderbeck:
Thank you so much for tackling this feature!

NodudeWasTaken:
Optimally, i would have opened the EditPanel upon creation, but that requires some redesigning of the panel and more extensive knowledge of how vuejs works.

Noob34567:
How is the progress on this? Create manual entries would be awesome and a good way to handle scrappers that don't work.

flonglong:
Nice! I guess this PR fixes https://github.com/xbapps/xbvr/issues/442 ? I'd love to have the capability to add manual scenes. I have a bunch of files that the scrapers for various reasons don't find.

PR #703, opened 2022-02-26
CHANGED: Added all available SLR Studios; Fix for duplicate SLR Studi…

Sniper-SnX:
…os having own scraperid;


theRealKLH:
A few things.. 
1. name formatting (minor).. you changed a few from their correct spelling (grin)... ex. it is "JimmyDraws" not "Jimmy Draws"/ "JustVR" not "Just VR".
2. you change existing slr scrapers.. ex. why add -slr to DDFNetworkVR? Scraping was moved from DDF site (dead) to SLR. this will dupe all scenes. 
3. wouldn't bother changing "dead" studio names.. ex. AltPorn4U to Markohs (they will [never say] NEVER release another video. It's been 4 years (2018) since the release of the existing 4 vids). you just create dupes for those of us who have this content already
4. overall.. any change to scraperID of existing scrapers is going to cause dupes....

Sniper-SnX:
1. The Idea behind is that it's not even consistent among different sites, JustVR is also found as Just-VR, etc. There are some like "VR Japanese Pornstars Stay Home Routines", "The Locked Cock Chronicles", and others which are too long to be displayed as one word. They tend to wrap in an ungood fashion in the admin ui. Maybe too ambitious to change even the small ones.
2. There exists an dedicated ddfnetwork scraper. As they share the same scraper id you cannot enable / disable this from within SLR hence an own id. IMHO duplicates doesn't really matter as there is always a convergence between original sites and an aggregator like SLR. Did this for all slr studio ids which have an dedicated own scraper (original site).

theRealKLH:
> Thanks for the feedback.
> 
np

> 2. There exists an dedicated ddfnetwork scraper.

not in xbvr there isn't (at least there shouldn't be).

Sniper-SnX:
So after a long time i updated the slr scraper
* Removed deprecated studies
* Tweaked available
* Aded new ones.
* Added Gay Studios
* Added Trans studios.

General notes here: 
* I used the spelling and naming from the slr site.
* All possible id collisions with existing scrapers will be avoided by adding a -slr suffix to the scraper id.

hiss-remi:
Still needs some work to properly handle trans studios. The check for whether sceneURL contains "scene" won't catch them, and then the scene scraper doesn't work because the scene pages don't contain the VideoObject json.

crwxaj:
This should be superseded by #1113

PR #716, opened 2022-03-18
fix_missing_previews

theRealKLH:
if funscript was added to scene before actual video, the preview was not being generated.


toshski:
I have found an issue with this fix that can cause XBVR to crash with the error
    **panic: runtime error: index out of range [1] with length 1**

GeneratePreviews now loops through files so it can skip funscripts. When ffmpeg sucessfully generates a preview it breaks out of the for loop and everything is fine. But if ffmpeg fails it loops around for the next file, if there are no more files, the panic is thrown as there is no index size check.
The following change on line 32 should fix it,

**for i < len(files) && files[i].Exists() {**

I can do a PR if you like

PR #718, opened 2022-03-21
projection_fix

### LINKED ISSUE #708: Projection Issues with DeoVR JSON
When using DeoVR, if a file is "matched", XBVR sends projection information (?) to DeoVR, and the ability to select a projection is **disabled** and hidden in DeoVR's UI. This is the same behavior it uses in "local playback" mode if a filename has something like "_mkx200" or "360" etc at the end of it - it picks that projection mode and hides the controls. Perfect example of how annoying that can be, even without XBVR entirely, is any JAV with "360" in its ID. ABC-360-R1.mp4 will lock to 360 degree TB - unless you rename it to something like ABC-36o-R1.mp4. **If the file is being played back from XBVR's "Unmatched" playlist, the projection information is freely selectable.**

~~XBVR seems to automatically handle all 360/TB videos perfectly. The sqlite database shows "360_tb", which doesn't seem to match DeoVR's documentation, but I guess XBVR is not only recognizing those videos, but properly sending whatever information it is DeoVR uses.~~

~~It doesn't seem to handle that at all for MKX200 videos. If I leave them unmatched, I can pick MKX200 just fine. I can also see that it recognizes them properly (somehow), since the sqlite database shows `mkx200` for those files under `video_projection` in the `files` table. I guess it's just not sending `"screenType": "mkx200"` to DeoVR?~~

~~Their documentation is confusing in general, this link https://deovr.com/app/doc#multiple-videos-deeplink shows different information compared to the "Web" one, this one mentions JSON files for browser usage which I assume is what XBVR does. The web one on the other hand talks about HTML and `<deovr>` elements.~~

It's worth mentioning that it looks like the projection type is set on a per-"scene" basis - so if a video has multiple files matched to it with different projections, then XBVR can only send one for the whole lot? e.g. a hypothetical "scene" with three videos, 1 regular SBS180equi, 1 TB360, and 1 MKX200 would have to "pick" a projection for all three, or alternatively (and preferably) have XBVR not send the projection information at all. Or whatever it's currently doing for "Unmatched" files that allows selecting your own projection instead of DeoVR forcing you into one by hiding the controls.

---

~~Whoops. After looking closer at the main.db file, the files with botched projections _didn't_ have mkx200 in the projection field, likely because it wasn't in the filename. After editing those entries, they work now.~~ Making that option visible somewhere in XBVR's UI would probably be preferable to having to edit an SQLite database for most people.

Nope - scratch that. Still seemingly busted for some files.

theRealKLH:
Should resolve #708 

This will allow selection of projection mode for scenes that have multiple videos in case they are different. Due to how DeoVR selects the same video from multi-file scenes each time a scene is played, the projection could be wrong, but it can now be changed.

### DIFF
--- a/pkg/api/deovr.go
+++ b/pkg/api/deovr.go
@@ -368,21 +368,23 @@ func (i DeoVRResource) getDeoScene(req *restful.Request, resp *restful.Response)
 		return cuepoints[i].TS < cuepoints[j].TS
 	})
 
-	if videoFiles[0].VideoProjection == "mkx200" ||
-		videoFiles[0].VideoProjection == "mkx220" ||
-		videoFiles[0].VideoProjection == "vrca220" {
-		stereoMode = "sbs"
-		screenType = videoFiles[0].VideoProjection
-	}
+	if len(videoFiles) == 1 {
+		if videoFiles[0].VideoProjection == "mkx200" ||
+			videoFiles[0].VideoProjection == "mkx220" ||
+			videoFiles[0].VideoProjection == "vrca220" {
+			stereoMode = "sbs"
+			screenType = videoFiles[0].VideoProjection
+		}
 
-	if videoFiles[0].VideoProjection == "180_sbs" {
-		stereoMode = "sbs"
-		screenType = "dome"
-	}
+		if videoFiles[0].VideoProjection == "180_sbs" {
+			stereoMode = "sbs"
+			screenType = "dome"
+		}
 
-	if videoFiles[0].VideoProjection == "360_tb" {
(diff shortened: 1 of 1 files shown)


crwxaj:
I think this should only be triggered if the assigned files have indeed different projections.

theRealKLH:
> I think this should only be triggered if the assigned files have indeed different projections.

It now works as intended. Only when different projections are detected for the scene. It may require reset of DeoVR cached data to work on existing scenes that were previously watched.

resolves #708 
resolves #726 (possibly a combination of DeoVR caching data)

PR #720, opened 2022-03-25
Add option to not track watch time

alexnsfw:


crwxaj:
What is the use case for this PR? Wouldn't it be enough to just disable DeoVR Remote?

alexnsfw:
Watch time is tracked even with remote mode disabled, it's just based on the time spent serving the file, and scenes still get marked as watched. I use this mainly to avoid scenes being marked as watched as soon as I open them to e.g. check the video quality, and I manage the watched state of scenes manually with #721.

PR #728, opened 2022-05-03
Heresphere changes

theRealKLH:
1. This exposes hsp files to HereSphere. If named correctly, will automatch same as videos & funscripts. Otherwise can be matched manually to the scene.
2. This fixes video duration in HereSphere "web stream" view.


theRealKLH:
reverted the change to fix video length display in HereSphere UI. Coding changes were done on HereSphere side to resolve.

PR #730, opened 2022-05-10
Fix VirtualRealPorn cast scraping

CoolMintChocolate:
VirtualRealPorn changed their cast page recently, this PR fixes that portion of the scraper.


crwxaj:
It looks like they changed it again, and the name is now in a h3 heading. Can you please check?

CoolMintChocolate:
On my computer it's still in a `<h1>`, this is a sample page I'm looking at: https://virtualrealporn.com/lady-lyne-porn/

I think you might be thinking of the models section at the bottom of the scene page. I'm pretty sure we don't get the actress names from there.

crwxaj:
Sorry, my mistake. I was indeed looking at the scene page.

PR #736, opened 2022-05-14
Tag and Actor count performance improvements

toshski:
Changed logic to return counts from the database using a group by query, rather than return scenes and them counting the scenes in code.
Counting Tags and Actors was taking approximately 5 mins on my system (27K scenes), this was quite frustrating when testing an issue with a scrapping and doing a lot of scraps.  But 5 mins at the end of a scrap is probably not a problem for most with normal usage, making this a nice to have change (it ain't broke). However given a drop from 5 minutes to a couple of seconds, may still be worth considering.

### DIFF
--- a/pkg/tasks/content.go
+++ b/pkg/tasks/content.go
@@ -352,27 +352,33 @@ func CountTags() {
 	var tags []models.Tag
 	db.Model(&models.Tag{}).Find(&tags)
 
-	for i := range tags {
-		var scenes []models.Scene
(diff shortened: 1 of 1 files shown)


crwxaj:
Good work, but this also counts tags and actors of deleted scenes. Deleted scenes remain in the database with a deleted_at field, that is NOT NULL.
Also, I don't think you need to group by id and count, the tag id should be enough.

toshski:
I'm use to MS Sql where not including count field would fail
Thanks for fixing the deleted records been included

PR #737, opened 2022-05-14
Create separate searchable fields for id, title, cast, site & description

toshski:
The current search process concatenates the scene-id, title, synopsis, cast and site into one field that is indexed so any of that information can be used for the basis of searches.
The change creates separate index fields for id, title, cast, site & description for indexing rather than a single one.  All separate fields are still searched by default, so you still get the same search results if you search for the same string as  you do now.
However, by splitting the fields into separate indexable fields, allows the opportunity specify which field you expect the search in in your query string and it will boost the score value for scenes where it matches that field. 
eg, with my data, searching for “Mommy Time”, it comes in at the 12th entry way down the quick find list and on page 3 when using the file match search.  Using the search in file matching, I can see the score for the scene is 66.91 both before and after my change.  However, by changing my search to “Title: Mommy Time” the scene I wanted jumps to the top with a score 591.33.  Note: it does not appear to stop searching of other fields, it just boasts the score if it matches the score if the field specified matches.
I still usually just search like now without specifying specific fields to boost, as the current search usually gets me what I want in the first couple of result over 90% of the time.  However, if I don’t get want I want, it has proved very valuable.
You can also specify multiple fields, eg “Title: Mommy Time Site:virtualtaboo Cast: Vittoria Dolce”
Impact on index space is fairly small, for 27k scenes I went from 89mb to 92mb

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "dev:ui": "cd ui && vue-cli-service build --watch --silent"
   },
   "dependencies": {
-    "@fortawesome/fontawesome-free": "5.15.3",
-    "@fortawesome/fontawesome-svg-core": "1.2.35",
+    "@fortawesome/fontawesome-free": "5.15.4",
+    "@fortawesome/fontawesome-svg-core": "1.2.36",
     "@mdi/font": "5.9.55",
     "buefy": "0.9.4",
     "bulma-extensions": "6.2.7",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "@fortawesome/fontawesome-free": "5.15.4",
     "@fortawesome/fontawesome-svg-core": "1.2.36",
     "@mdi/font": "5.9.55",
-    "buefy": "0.9.4",
+    "buefy": "0.9.20",
     "bulma-extensions": "6.2.7",
     "date-fns": "2.16.1",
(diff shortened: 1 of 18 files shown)


crwxaj:
Does this work on the fly or do we need to migrate the search index somehow?

toshski:
Good point, I'll do some checks and see if it just works but I suspect I'll need to work out how to migrate it

toshski:
The indexes don't migrate. Searches do continue to work with the exact results they use to, very surprised given they are indexed under the field "fulltext" which is no longer referenced.  The indexes do still have to be re-indexed though,  if someone uses a new prefix, "titile:" etc, then I expect new scenes will be boosted up which are indexed under the new fields and old scenes will actually get pushed down.  
I tested a  change to migrate the indexes. It uses the data migrations in migrations.go,  I realize that is normally intended for data, but it seems to work fine.  If your happy with that approach I can commit it.

crwxaj:
could you please update the PR based on the new release version, so I can merge it for the next release.

toshski:
Sorry, I think I've mucked something up in my environment. I might close this PR, sort my environment and resubmit as a new clean PR

toshski:
Replaced by PR 760

PR #738, opened 2022-05-14
configure buttons in scene list

crwxaj:
This PR makes all buttons in the scene list configurable.
Sadly, these options are only loaded when opening the options page, so it still shows only default buttons on initial site load.
I would appreciate if anyone with Vue.js skills can have a look at it and make the optionsWeb state load on startup.


crwxaj:
Hey @cld9x could you please contact me in Discord.

PR #755, opened 2022-06-13
Backup restore v2

toshski:
The current backup, exports data by scrapping the web sites.  This has some draw backs

- Changes to the site breaking scraping, means you can't export data
- You cannot export user data, such as favourites, star ratings values, Cuepoints, etc
- Populating test environments is slow using scraping

A standard database backup is probably all a lot of people may need, however the new function provides a finer level of control when required and this process can backup/restore between Sqlite and MySql/MariaDb

The new Backup/Restore is based on the existing Import/Export but sources data from the database.

It is flexible with options to Backup/Restore

- Use the new or the existing json format file with the new format supporting more data fields and entities,
- Only New Records or to New and Overwrite Existing data
- Storage Paths
- Saved Searches
- Scene Data with separate options for
   - Scene master data
   - Scene/File links (to save file links made manually)
   - Cue points
   - Watch History
   - Edits (actions table)
   - Choose All Sites or only Enabled Sites for the scene related data options

The new output format is based on the structs used for the database rather than the ScrapedScene which enables data that wasn't scraped, e.g. Favourite Flag, Star Ratings, manually added tags etc., and also non-scene data i.e. Saved Search definitions and Storage Paths

It does use a new Github package json-iterator/go to specify new tags to control the json, the existing json tags on the structures are used by internal processes but are not always what I needed for the json bundle file.

Further background is at Discussion [https://github.com/xbapps/xbvr/discussions/748](https://github.com/xbapps/xbvr/discussions/748)


toshski:
Added an option for the user to select a Saved Search to filter scenes outputted in the Backup

toshski:
Updated to resolve merge conflicts with other commits

theRealKLH:
question on function... did a restore and no sites were enabled. is site status backed up?
image

toshski:
> did a restore and no sites were enabled. is site status backed up

Yes.  If you check the bundle file you are restoring you should have a json entry similar to below

>  "sites": [
>   {
>    "name": "100%VR (SLR)",
>    "is_enabled": true
>   },

toshski:
I have made the requested changes. Some notes

- I have removed the Version number field, however, people can still upload a version 1 file, so it won't break any imports people may have built for xbvr
- I have used Tooltips to provide extra notes/info without cluttering up the screen
- The download, will create a file on the server, in the appdir/downloads directory.  I tried streaming the download data directory back through the api call.  This worked fine.  But with further testing with 20k scenes, Chrome was fine, but had problems using Firefox and Edge. Although Edge was fine, if it wasn't set to prompt Save As (weird).  Creating a file and using a download link to the file rather than api url proved a lot more robust.

UI is probably not my strength not if there are any suggested changes let me know

crwxaj:
Can you please update to the current master branch.
Also I found a lot of primary keys in the bundle, which are specific to the current DB, and not actually used. Can you remove them from the export?

toshski:
Updated to current master branch
Removed internal database ids from export output

PR #759, opened 2022-06-17
Fix cover breaking when editing scene

alexnsfw:
Just reverting a small change introduced in the last release that causes the cover to be removed from the scene card whenever a scene is edited.


flonglong:
Thanks!

PR #767, opened 2022-06-21
fix_incomplete_tngf_migration

theRealKLH:
fix semi BAD migration


pkg/migrations/migrations.go
```diff
@@ -726,13 +726,13 @@ func Migrate() {
...
...ogether-with-her-longtime-fan-26720"},{"date":"2021-09-09","sceneid":"tonight-s-girlfriend-vr-29338","sceneurl":"https://www.tonightsgirlfriend.com/scene/scarlett-mae-shows-up-looking-delicious-and-sexy-in-lingerie-for-her-client-29338"},{"date":"2021-09-23","sceneid":"tonight-s-girlfriend-vr-29976","sceneurl":"https://www.tonightsgirlfriend.com/scene/penelope-kay-takes-care-of-her-big-dick-client-29976"},{"date":"2021-10-07","sceneid":"tonight-s-girlfriend-vr-30114","sceneurl":"https://www.tonightsgirlfriend.com/scene/diana-grace-dresses-in-sexy-red-lingerie-for-her-big-cock-client-30114"},{"date":"2021-10-21","sceneid":"tonight-s-girlfriend-vr-30629","sceneurl":"https://www.tonightsgirlfriend.com/scene/the-sexy-petite-pornstar-brooklyn-gray-shows-she-can-take-a-big-black-cock-anytime-30629"},{"date":"2021-11-04","sceneid":"tonight-s-girlfriend-vr-30648","sceneurl":"https://www.tonightsgirlfriend.com/scene/the-sexy-blake-blossom-puts-on-special-lingerie-just-for-her-client-30648"},{"date":"2021-11-19","sceneid":"tonight-s-girlfriend-vr-30677","sceneurl":"https://www.tonightsgirlfriend.com/scene/gorgeous-and-fit-babe-ana-foxxx-takes-care-of-her-fan-in-every-way-he-desires-30677"}]`
```
alexnsfw, line 735: Why are there fewer scenes than before?

pkg/migrations/migrations.go
```diff
@@ -747,33 +747,37 @@ func Migrate() {
...
 							scene.SceneURL = scenejson.Get("sceneurl").String()
 
 							// update all actions referring to this scene by its scene_id
-							err = tx.Model(&models.Action{}).Where("scene_id = ?", scene.SceneID).Update("scene_id", scene.SceneID).Update("release_date", scene.ReleaseDate).Update("release_date_text", scene.ReleaseDateText).Update("scene_url", scene.SceneURL).Error
+							err = tx.Model(&models.Action{}).Where("scene_id = ?", scene.SceneID).Update("scene_id", sceneID).Error
 							if err != nil {
 								return err
 							}
-
+							scene.SceneID = sceneID
```
alexnsfw, line 774: This will lead to duplicate scenes even sharing the same scene id if you've run the new scraper, right?

crwxaj:
Can you please have a look at Alex' concerns. The way I see it this will indeed create duplicate scene ids if you used the native scraper in the meanwhile.

pkg/migrations/migrations.go
theRealKLH: Ana Foxxx's scene is the last to be migrated

theRealKLH:
same sceneid SHOULD overwrite. The dupes would be due to same scene with different ids

theRealKLH:
Ok if someone wants to check my sanity (why would I think sceneid is unique? duh!) this should do it. 

1. removes "native" scenes if no video is associated (video is not "available" or "accessible").  
2. completes migration from vrporn.
3. scene count may not be 100%. Current count is 49. first scrape should add missing scenes.

PR #849, opened 2022-07-13
Split scene status update from storage rescan

### LINKED ISSUE #340: Option to Rescan select folders
I have current download folder which has new scenes

Currently the app scans all directories
the only way to scan select directories is to disable hard disk / make it offline
does not work in case I want to disable select partition

does not make sense when most of the data is not changed

please add option to scan 'select locations' in addition to 'Rescan All'

rescan

### LINKED ISSUE #691: Rescan Process Performance
Hello,
I'm wondering if the process of adding new files could not be more efficient, esp. since I'm running the application on a NAS which doesn't have the greatest performance. 

So, when I add a new file, I go to Storage -> Rescan and then:
a) all existing folders are scanned -> here a possibilty would be great to only trigger the rescan for a specific folder (this is already an old open issue: #340 )

b) the "Update status of Scenes" process takes quite long -> somehow for this all of the existing scenes (in my case almost 20k) are processed -> is there a reason why this is necessary each time, or would it be possible to only process newly addes files / scenes? 

c) the "Tag Count" process takes very long -> would it be possible to disable the tag count feature and this processing in general? Or at least make it more efficient by only adding the tags from newly added files / scenes, instead of going through ALL scenes and completly count again from zero each time (which seems to be the process right now)

toshski:
resolves #691 
resolves #340 

This change splits out the Scene Status update from the Storage Scan.  A new button has been added so the user can manually choose to run the Scene Status update when they are ready to.

A scene status update is run on specific scenes added by the Storage Scan, so they will be up to date
The Scene Status update will still run after the Storage Scan when triggered by the Cron task every 2 hours. 

There is also the ability added to scan a single storage path

### DIFF
--- a/pkg/models/model_volume.go
+++ b/pkg/models/model_volume.go
@@ -101,8 +101,23 @@ func CheckVolumes() {
 	var vol []Volume
 	db.Find(&vol)
 
+	var files []File
 	for i := range vol {
-		vol[i].IsAvailable = vol[i].IsMounted()
-		vol[i].Save()
+		isMounted := vol[i].IsMounted()
+		if isMounted != vol[i].IsAvailable {
+			vol[i].IsAvailable = vol[i].IsMounted()
+			vol[i].Save()
+
+			// update the status of any scene with a file on that volume
+			db.
+				Model(&files).
+				Where("volume_id = ?", vol[i].ID).
+				Find(&files)
+			for _, file := range files {
+				var scene Scene
+				scene.GetIfExistByPK(file.SceneID)
+				scene.UpdateStatus()
+			}
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


crwxaj:
I think, if you also add an UpdateScene for all deleted files, we can get rid of scheduled full status updates altogether, and leave the full status update only as a means to fix the status manually.

toshski:
Can I check where you are thinking of.  I've retested the following functions that I can think of as relevant and they seem ok.
- Deleting and Un-matching from the FIles/Matched Tab
- Deleting and Un-matching the file from the Scene Details (needs a refresh when returning to the Scene List, but status is ok)
- Deleting a Path from Options/Storage 

If you are thinking of the Deleting a path from the Storage screen, it does a full refresh of scenes status.  I have added that as part of the change. 

Was there somewhere else you were thinking of. If not I will add a commit to just remove the RefreshSceneStatuses from the scheduled rescan.

PR #870, opened 2022-07-21
title_size_change

theRealKLH:
Changes size of the title column for mysql. sqlite3 is not modified (only new installs reflect change). sqlte3 does not obey size restrictions by default unless table was created as "STRICT".


theRealKLH:
> Is the call to tx.AutoMigrate() neccessary in the migration?

It was in there before I added mysql condition and before I realized that sqlite ignores whatever is set for the column anyway.

crwxaj:
AutoMigrate failed on my SQLite with error "Could not migrate: duplicate column name: created_at". I will remove it.

PR #871, opened 2022-07-21
equirantangular_support

theRealKLH:
Adds support for filename modifiers _180 & _360 for non SBS/TB VR (equiranctanglular cubemap files not included _360EAC & _EAC360)


crwxaj:
Won't this also set "TB_360", "LR_180" and "180_180x180_3dh_" files to the non-stereo mode?
Please make the detection way more specific, since mono files should be a very rare edge case.

theRealKLH:
DAMN YOU!.. grin... this is why it's good to have a second pair of eyes. I shortened my original testing...

PR #890, opened 2022-07-24
multi-versions

theRealKLH:
Adds button to filter scenes that have multiple video files associated. It does not include scenes that are identified as "multipart" (is this used outside of javr videos?). The idea is for scenes that have multiple versions (4k, 7k... 264,265... 180,360.. etc.)..


theRealKLH:
superceded by #1063

PR #891, opened 2022-07-24
Fix All Sites Switch in Import/Export

toshski:
The following in the OPtion Import/Export page  `this.allSites == "true"`  always evaluates as false.  This means the All Sites switch isn't working and only processes enabled sites.

If a user selects All Sites, they maybe missing scenes and be unaware of the issue


crwxaj:
I think `this.allSites == "true"` is correct, since radio controls have strings as values, while switch controls are mapped to boolean. It still works if you change it once. I think the real problem is, that allSites is still initialized as a bool value in line 157

toshski:
Ok, I wondered why you changed it, that makes sense now and why the others are still the same.  I update the PR to change line 157 and put it back.  First test look ok

PR #898, opened 2022-07-29
patch-fisheye180

vt-idiot:
* fisheye_projection_type

Added 180 degree fisheye lens type (detected by filename _fisheye)

I looked at #807 and tried to imitate it. DeoVR API does claim to separately support 180 degree fisheye projection, [9. Naming convention](https://deovr.com/app/doc#naming) & [11. Supported type of meshes](https://deovr.com/app/doc#Supportedmeshes)

I am not sure if this will break parsing for fisheye vs. fisheye190. A considerable chunk of older JAV uses linear 180 degree fisheye and not mkx200.

### DIFF
--- a/pkg/tasks/volume.go
+++ b/pkg/tasks/volume.go
@@ -217,7 +217,7 @@ func scanLocalVolume(vol models.Volume, db *gorm.DB, tlog *logrus.Entry) {
 							if part == "mkx200" || part == "mkx220" || part == "rf52" || part == "fisheye190" || part == "vrca220" || part == "flat" {
 								fl.VideoProjection = part
 								break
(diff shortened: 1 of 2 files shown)


pkg/tasks/volume.go
```diff
@@ -217,6 +217,9 @@ func scanLocalVolume(vol models.Volume, db *gorm.DB, tlog *logrus.Entry) {
 							if part == "mkx200" || part == "mkx220" || part == "rf52" || part == "fisheye190" || part == "vrca220" || part == "flat" {
 								fl.VideoProjection = part
 								break
+							} else if part == "fisheye" || part == "F180" || part == "180F" {
```
crwxaj, line 220: The file name is converted to lower case, so the capital F won't match

PR #906, opened 2022-08-11
support for HereSphere native API

crwxaj:
Since HereSphere allows you to namespace tags, I add some feature tags to each scene, like "Feature:Cast: 2", "Feature:Resolution: 8K" or "Feature:Has HSP file". This might not be the most reliable feature, since tags only get updated when you open the scene, or do a full rescan, but it might come in handy.

### DIFF
--- a/pkg/api/heresphere.go
+++ b/pkg/api/heresphere.go
@@ -82,12 +82,11 @@ type HereSphereAuthRequest struct {
 func HeresphereAuthFilter(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
 	if isDeoAuthEnabled() {
 		var authorized bool
-
 		var requestData HereSphereAuthRequest
-		if err := json.NewDecoder(req.Request.Body).Decode(&requestData); err == nil {
+
+		if err := json.NewDecoder(req.Request.Body).Decode(&requestData); err != nil {
 			authorized = false
 		} else {
-			log.Infof("%+v", requestData)
 			err := bcrypt.CompareHashAndPassword([]byte(config.Config.Interfaces.DeoVR.Password), []byte(requestData.Password))
 			if requestData.Username == config.Config.Interfaces.DeoVR.Username && err == nil {
 				authorized = true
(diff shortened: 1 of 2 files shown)


vt-idiot:
How difficult would it be to extend this to subtitle detection? HereSphere recently added support for subtitles, they play automatically if there is an .srt file in the same directory and with the same filename as the video. AFAIK XBVR just ignores the .srt files when it scans.

crwxaj:
It shouldn't be that hard to implement this similar to funscript and hsp files. Where can I find subtitled scenes for testing?

vt-idiot:
I'm honestly not aware of any outside of JAV scenes, and I only have one file handy that is pretty sparse with lines and timing.

Here's a random .srt file that starts at 00:00 and is heavy enough on dialogue that you'll quickly be able to tell if it is working. If HereSphere can see it, the closed captioning "**CC**" icon will allow you to enable subtitles. Just rename it to the same exact filename as the scene you're testing with with an `.srt` extension. e.g. `Some-Studio-Some-Scene-5K-HEVC.mp4` should have `Some-Studio-Some-Scene-5K-HEVC.srt` in the same directory. If (without XBVR) you use HereSphere to navigate to that folder and play the video, it will automatically detect the subs. I don't use scripts but I'm pretty sure it looks for those filenames the same way.

[rename me and change my extension to .srt.txt](https://github.com/xbapps/xbvr/files/9409659/rename.me.and.change.my.extension.to.srt.txt)

Speaking of JAV, they're all multi-part scenes. I know my way around ffmpeg/mp4box and things like Subtitle Edit well enough to concatenate the scene and the subtitles if I absolutely _have_ to, but I'm curious if both the HereSphere API and XBVR itself can handle subtitles (or scripts) for a multipart scene. e.g. a hypothetical scene ID JAV-1234 with two parts, both matched, and files like this:

> `JAVID-1234-A.mp4` + `JAVID-1234-A.funscript` + `JAVID-1234-A.srt`
> `JAVID-1234-B.mp4` + `JAVID-1234-B.funscript` + `JAVID-1234-B.srt`

crwxaj:
I found some subtitled scenes on CzechVR: https://www.czechvrnetwork.com/tag-eng-subtitles
Alas, apparently subtitles don't work yet on the Quest version of HereSphere, so I won't be able to test it. I'm afraid, I won't be able to tackle that issue until subtitles are supported by the Quest 2 port.

As far as I know, no VR player so far has a concept of multipart-scenes, and I don't watch JAV myself, so I don't know what is the best way to handle that. Currently the only thing I can think of is creating multiple custom scenes in XBVR.

PR #917, opened 2022-08-16
minor: fix menu entry in tray app

crwxaj:


theRealKLH:
my bad. looks like I forgot to copy that line from the original PR before I closed it.

PR #943, opened 2022-08-31
OShash support for files

cld9x:


crwxaj:
The PR looks fine to me, but I don't see what use have OSHashes for XBVR.

cld9x:
> The PR looks fine to me, but I don't see what use have OSHashes for XBVR.

stash-db and tpdb are using it for querying

PR #950, opened 2022-09-11
Add ability to schedule Preview Generation

toshski:
Added the ability to schedule the running of Preview Generation.

Now that schedule tasks can have a start windows, users with XBVR running in low powered environments, can use the "Limit time of day" option to specify a period of low use to generate the Previews, so the resource intensive process is not running when they would normally use the system.  The schedule task "limit time" option usually just controls the time a job can start.  GeneratePreviews has been modified to stop generating Previews when the end of the time range for the schedule task is reached.


crwxaj:
The endTime calculation is wrong, when the time frame passes midnight. I set it to run from 20:00 to 02:00 and it set the endTime to 02:58 today (not tomorrow) and thus the preview generation doesn't even start.

toshski:
Thanks, not sure how I missed that as the midnight change took some testing.  I also noticed a case where if the job was scheduled to start x minutes pasted the hour, you could also not get the job run

Issue #955, opened 2022-09-22
R18 is shutting down, JAVR scraped scenes need a migration


vt-idiot:
Which means images, cover_url, and scene_url for any JAVR scraped scenes will break.

![www r18 com_](https://user-images.githubusercontent.com/81622808/191868256-6f643299-b566-4b38-8cb8-cd7820ff00a1.png)

FANZA is not shutting down in Japan, and their URLs for images are identical after the first slash - I was able to test the following manually on my database, and it seems like all of the images still load:

```sql
UPDATE scenes
SET
	images = REPLACE(images, 'https://pics.r18.com/', 'https://pics.dmm.co.jp/'),
	cover_url = REPLACE(cover_url, 'https://pics.r18.com/', 'https://pics.dmm.co.jp/');
```

Alternatively, if you also want to change the links to the scenes themselves, although they will be entirely in Japanese:

```sql
UPDATE scenes
SET
	images = REPLACE(images, 'https://pics.r18.com/', 'https://pics.dmm.co.jp/'),
	cover_url = REPLACE(cover_url, 'https://pics.r18.com/', 'https://pics.dmm.co.jp/'),
	scene_url = REPLACE(scene_url, 'https://www.r18.com/videos/vod/movies/detail/-/', 'https://www.dmm.co.jp/digital/videoa/-/detail/=/c');
```

That extra line will change a link to e.g. `https://www.r18.com/videos/vod/movies/detail/-/id=wavr00066/` to `https://www.dmm.co.jp/digital/videoa/-/detail/=/cid=wavr00066/`

I am not 100% sure how migrations work, but I'll open a pull request for one that tries to do all of the above.

vt-idiot:
Resolved with #956

PR #956, opened 2022-09-22
JAVR Migration from R18 URLs to Japanese FANZA URLs

vt-idiot:
R18 is closing down. Proposing Migration ~~0042~~ 0044 to fix images.

See #955 

I'm not 100% sure I've done this properly, as I just looked at migration 0035 and migration 0036 and tried to copy what they did, but for this scenario.

Ideally you'd want to select only for scenes that have "JAVR" as a tag, since the scraper always added that, but that doesn't exist under "scenes" in the database. I think it ultimately doesn't matter in the end since **nothing** else will be selected against `www.r18.com` or `pics.r18.com`? Unless there's a performance consideration to be made for a migration that happens in this manner, but doing it manually via SQL only took 3 seconds:

```SQL
Execution finished without errors.
Result: query executed successfully. Took 2803ms, 20418 rows affected
At line 1:
UPDATE scenes
SET
	images = REPLACE(images, 'https://pics.r18.com/', 'https://pics.dmm.co.jp/'),
	cover_url = REPLACE(cover_url, 'https://pics.r18.com/', 'https://pics.dmm.co.jp/');
```

The only thing I am not sure about is if "CoverURL" and "SceneURL" can be changed in this manner, or if it has to look more like migration 0036.

I left the "SceneURL" change commented out because I'm honestly not sure what to do about that yet. Come January, or sooner if they pull the plug before then, those SceneURL links **will** be dead.


vt-idiot:
gofmt isn't complaining anymore, I think the issue was the block comment
not sure if the migration itself would work properly though - reviews welcome...

vt-idiot:
Fixed the conflict with migrations 42 and 43, updated the first comment I opened the PR with a bit to demonstrate what I am trying to do if I've somehow mucked up the migration code.

vt-idiot:
Done. The R18 links already stopped working. The only thing I'm still not certain on is if this Go migration will behave exactly the same way as the SQL I posted above, because I just "monkey see, monkey do"-d another migration into this. ie. will it do this?

- Scene URL: `https://www.r18.com/videos/vod/movies/detail/-/id=kavr00015/` -> `https://www.dmm.co.jp/digital/videoa/-/detail/=/cid=kavr00015/`
- Cover URL: `https://pics.r18.com/digital/video/kavr00015/kavr00015pl.jpg` -> `https://pics.dmm.co.jp/digital/video/kavr00015/kavr00015pl.jpg`
- Non-Cover Image URLs: `https://pics.r18.com/digital/video/kavr00015/kavr00015jp-1.jpg` -> `https://pics.dmm.co.jp/digital/video/kavr00015/kavr00015jp-1.jpg`

If the code looks like that's what it does, then hooray.

crwxaj:
The code looks fine to me.

PR #985, opened 2022-10-15
Treat issues decoding Heresphere Body as a Warning

toshski:
Currently if there is an issue decoding Body of a Heresphere request then the request is ignored (error) and no scene data is returned to Heresphere.  This change will still log a warning but will still return the scene data to heresphere.


toshski:
Have seen some errors after the change, closing until I recheck

PR #1020, opened 2022-11-15
Trailer Playlist

toshski:
New Feature to enable users to create a list of scenes that they wish to view the Publishers Trailer.  Trailers can then be viewed using Heresphere.  This is faster and easily than going to individual sites and dealing with different players.

A new Trailer List button has been added for the user to indicate they wish to watch a trailer for the scene.
image

The Trailer List button is only enabled for scenes that don't have a downloaded file
 
If the user has added any scenes to the Trailer Playlist, then a Trailers Tab will appear at the end of the Saved Search Tabs (if the Trailer Playlist is empty this tab won't appear).

A lot of publisher's trailers URLs include attributes that make the URL unquiet, eg the viewer IP address, and valid to time.  Therefore, the trailer URLs usually need to be generated when the user is ready to view them.  When you go the Trailer Tab in Heresphere, it will then request the scene details and the url will be generated at that point.  The URL are generated based on about 6 different methods. The rules for a scene are setup when the scene is scrapped and saved in the scenes record.

A few old sites that have not had scenes in the last year have not been setup.

Sites scrapped from SLR have not been included.  I have coded and test watching SLR trailers using 2 methods, using the SLR Api and scraping json from the scene pages and these sites work well.  However, I opted to exclude them given SLR attitude towards downloads and Heresphere.  This could easily be added back if it isn't actually a problem to include them, see code in commit 7cd3aab which disabled it.  I have not tried to implement this with the DEOVR player for the same reason, although I expect there is no reason it could not be implemented as well as the api is very similar.

There are 3 ways to remove a scene from the Trailer List, the XBVR web UI, while in Heresphere, go the Video Tags and remove the "Feature: Trailer List" tag (my preferred way) or by downloading the scene (this actually still leaves it in the Trailer List but only scenes with no video file are sent to Heresphere.)

Migrations exist to populate the Trailer Url rules for old scenes, it may take a few minutes for the migration to populate the old scenes.

I typically will scrap for new scenes once a day, then I add scenes I might be interested into the Trailer Playlist and view them in Heresphere.  If I decide I want to download the full scene, I add a tag "Category:download" (required Tag Updates to be turned on in the API). This creates a tag "download" on the xbvr scene, and I can then easily filter "download" tags in xbvr to get the list of scenes to download. Heresphere remember what Tags you recently used, so it make that really easy to add the Category:download tag.

Sites vary in the their ability to stream trailers, some have one, some have multiple resolutions, some have h264 and h265.  Some don't play well on their higher bit rates and you may need to drop down.  Jumping forward in clips is often a poor experience, given the clips are usually only a couple of minutes long, I would recommend just playing through.

Sorry about the long description 😄


crwxaj:
Great work, but the migration crashed for me, so I added a few safe guards. Please have a look.

toshski:
And the migration should have been the easy bit.  Should get time tomorrow, but I'll have to reset up a new test data.  You didn't get a line number of message by chance.

crwxaj:
I already fixed my issues in the latest commit.

First I got `Could not migrate: duplicate column name: created_at`, so I moved the schema migration to a separate step and only added the new fields.

The next issue was `panic: runtime error: index out of range [1] with length 0`, because baseURL[1] wasn't matched, so now I skip the scene if len(baseURL) == 0.

Also I added some array length checks for the other URL splits.

Please have a look if my changes may break anything.

toshski:
Last 2 commits to address merge conflict and formatting

PR #1047, opened 2022-12-04
size reps

theRealKLH:
amount representations "fixed" in Files tab. Mbit instead of MB and GiB instead of GB


crwxaj:
As discussed, there is no real standard here, and personally I like metric GB better than binary GiB, because of easier conversions.

PR #1048, opened 2022-12-05
Cuepoint Filtering Enhancements

toshski:
Adds a number of enhancements to Cuepoint searches with the Scene list filtering

- Added an Index to scene-id in the scene_cuepoints table.  This was primary for searching for all scenes with no Cuepoints (search for must not include %).  On a system with a large number of Cuepoints, the UI would timeout before the backend would finish querying scenes without cuepoints.
- Allow an * (astrix) wildcard to be used. Cuepoint filtering uses like clause and therefore a % wildcard can be used in searchs.  However, * is more commonly recognized as a wildcard by people who don't know sql.  This change treats * as a % and allows both to be used
- Allow Double Quotes to disable wrapping wildcards around search terms.  The current search wraps %'s around the search term, so doggy will also find doggystyle.  However, this means you can't just search doggy.  With this change, if the user wraps the search string in double quotes, then the like clause will not wrap the search term with %, e.g., "doggy" will only find Cuepoints that match doggy exactly.
- Refactor code to generate Cuepoint query, to avoid possible SQL injection type issues


toshski:
All fine, I saw an issue caused by bad data from testing another mod

PR #1049, opened 2022-12-05
Scene Property Filtering Enhancements

toshski:
This feature adds some enhancements to scene filtering using the fields under "Properties", i.e., Watchlist, Scripted, etc.

- Changed the buttons to have 3 states for searching, similar to what was added to other filters in the previous release, i.e., you can search any, must have and must exclude, e.g., any scene, scenes with scripts, scenes without scripts. The new state been must exclude.
- Added Trailer list as a new Property, (requested by a user)
- Added Ratings as a new Property, i.e., filtering scenes with Ratings or scenes without Ratings is now possible

Technical Details:
The three states are recognized by Transparent-undefined (Any), Green-true (must include) and Red-false (must exclude).  
Unfortunately, I couldn't find a way to toggle a Red color when the b-button-checkbox was false.  In the end I had to change to using a normal button, but this resulted in more code to do what the previous b-checkbox-button did for us.  If there is a way to toggle color of a b-checkbox-button, the code change could be simplified

### DIFF
--- a/pkg/models/model_scene.go
+++ b/pkg/models/model_scene.go
@@ -538,14 +538,26 @@ func QueryScenes(r RequestSceneList, enablePreload bool) ResponseSceneList {
 	}
 
 	for _, i := range r.Lists {
-		if i.OrElse("") == "watchlist" {
-			tx = tx.Where("watchlist = ?", true)
-		}
-		if i.OrElse("") == "favourite" {
-			tx = tx.Where("favourite = ?", true)
-		}
-		if i.OrElse("") == "scripted" {
-			tx = tx.Where("is_scripted = ?", true)
+		truefalse := true
+		fieldName := i.OrElse("")
+
+		if fieldName != "" {
+			if strings.HasPrefix(fieldName, "!") { // ! prefix indicate NOT filtering
+				truefalse = false
+				fieldName = fieldName[1:]
+			}
+			switch fieldName {
+			case "star_rating":
+				if truefalse {
+					tx = tx.Where("star_rating > ?", 0)
+				} else {
+					tx = tx.Where("star_rating = ?", 0)
+				}
+			case "scripted":
+				tx = tx.Where("is_scripted = ?", truefalse)
+			default:
+				tx = tx.Where(fieldName+" = ?", truefalse)
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


toshski:
I need to retest this, I may have broken something in the final code cleanup

toshski:
All fine, I saw an issue caused by bad data from testing another mod

theRealKLH:
Not sure I'm "liking" the functionality here. In my VERY limited test, it requires too many clicks to change filtering options. If I want "Favorites", I click once (green). Then if I want "Scripted", I have to click "Favorites" twice, and "Scripted" once (green).

FYI.. not sure it's "supposed" to work the way it is currently, but if you use the "sort by" dropdown on ratings, it magically hides all scenes that aren't rated.

toshski:
That scenario sounds like you are wanting to clear a previous selection and do a new one.  You could always just choose Default from the drop down which will reset to the default.  Currently you want "Favorites", you click once on Favorites. Then if you want "Scripted", I have to click "Favorites" once, and "Scripted" once.  So, you are only talking one extra click.  I have found I've been using the Not (Red) settings a lot more than I ever though, so personally I would take that trade-off for functionality that I had no way of doing.

In terms of the Ratings "sort by", pretty sure that's always been the case, if it's not rated it's not included.  No way to get a list of scenes not rated.  Same with other options such as Last Viewed, if you haven't watched it once, it wouldn't appear

theRealKLH:
So for me, after thinking a bit more... (I actually deleted the half page paragraph reiterating why I don't like it-GRIN)...

I can get behind this now (grin) HOWEVER.... I guess what we REALLY need (if we are going multistate on "filters") is a reset button to clear each "filter" (like month & folder). Yes the "default" pulldown does the same thing (thanks, that wasn't in my thinking), but how many people you think are using that as a "reset" (myself included-heh)? It IS labeled "saved searches" after all. Also the fact that it will remove ALL filters, and maybe I don't want that nuclear option (example, might want to keep filtered cast/tags/sort).

toshski:
Good suggestion.  
I do quite like the idea of a nuclear option of a simple one button click to reset all as well, but another button may start to clutter up the UI. So, I'll just stick to your suggestion, given you can still clear all with the Default.

toshski:
Added a suggestion to #1066 to make Watchlist/Favorites/Scripted filterable Attributes as part of that mod.  That way people still do filtering such as Not Scripted.  We can then leave the Property buttons as they are and drop this mod.

crwxaj:
superseded by #1066

PR #1055, opened 2022-12-13
Add ability to sort the file lists sent player Apis

toshski:
This change allows users to specify the sequence of files sent to the players in the APIs.

It is I response to an issue a user was having with a Multipart scene with a script for each.  The script files were in a strange order, the existing sequence is Selected Script, then Date Created.  They really just need by filename. As a work around they had to edit and resave the script files to achieve the order they needed.

I opted to make the solution more generic, rather than just by Filename, just for Script files. This was because I was doing a change anyway and maybe that wouldn't work for everyone. 

I have allowed for separate control of the sequence of both Video and Script files.  Users can choose one or more field, and each can be Ascending or Descending
Script files can be sorted by Filename, Added, Updated and Selected
Video Files can be sorted by Filename, Added, Updated, Resolution, File Size, Bit Rate, Frame Rate, Codec, Duration. 

Fields can be selected under Options/Players/Shared Player Options

The change has been applied to both Heresphere and DeoVR APIs

### DIFF
--- a/pkg/api/deovr.go
+++ b/pkg/api/deovr.go
@@ -320,7 +320,7 @@ func (i DeoVRResource) getDeoScene(req *restful.Request, resp *restful.Response)
 	var sources []DeoSceneEncoding
 	var sourcesSpatial []DeoSceneEncoding
 	var videoFiles []models.File
-	videoFiles, err = scene.GetVideoFiles()
+	videoFiles, err = scene.GetVideoFilesSorted(config.Config.Interfaces.Players.VideoSortSeq)
 	if err != nil {
 		log.Error(err)
 		return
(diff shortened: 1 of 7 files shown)


crwxaj:
The "selected script" feature was added deliberately, since DeoVR doesn't support alternate scripts, so you have to choose one in the Web UI.
Could you please make sure, that XBVR preserves the current default, if someone doesn't use this feature?

toshski:
I wondered what it was for.  I'll have a look and sort the new Merge conflicts

toshski:
I've fix an issue introduced a couple of merge conflicts back. This is the change in commit [90a49c2](https://github.com/xbapps/xbvr/pull/1055/commits/90a49c223f973346088419ddf950c9e01fc033e9)

But I think I've screwed something up with the last commit 485d1de, I think I merged my test branch into this one by mistake.  Is there anyway to undo 485d1de in git. I can undo it in my dev environment, but I'm not sure if that will let me push it again as it will be out of step.  Alternatively, I could fix in my environment and push it as new PR

PR #1057, opened 2022-12-13
VRPorn Scrapper Fixes

toshski:
The VRPorn scrapper stopped scrapping scenes a couple of weeks ago.  
The sites scrapped from VRPorn have not had any new scenes for a while, so there is no urgency with this.  Mod is in case any do, or if new sites are added in the future.  I only noticed as my fork has a couple of other sites from VRPorn. 

- The tax-studio class has shifted from the post's HTML article tag to the body tag
- Removed evileyevr as it is no longer available on VRPorn

### DIFF
--- a/pkg/scrape/vrporn.go
+++ b/pkg/scrape/vrporn.go
@@ -126,7 +126,7 @@ func VRPorn(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out chan<
 		}
 	})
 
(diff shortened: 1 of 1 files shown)


theRealKLH:
@toshski We should probably change the collector to (sort by newest)

`siteCollector.Visit("https://vrporn.com/studio/" + scraperID + "/?sort=newest")`

PR #1058, opened 2022-12-13
Add Members_url to scene

toshski:
A number of sites have a separate member's area with no links or redirects to easily go between their public facing area and their members area.

To download content you have to search for the scene first.  This feature adds a Members Link so you can go straight to the scene in the members are.
Sites included in the new feature are

- All Czech VR Network sites
- NaughtyAmeric VR
- VirtualPorn
- WetVr
- VRHush

Member urls are not created for sites that don't benefit from the feature, i.e., they may redirect or if logged you get the member features anyway.

Note: VRHush is based on information provided, I don't have a membership, so the links are untested.

This feature in no way negates needing a valid membership and to be logged in.

Some users are redirecting across domains to other aggregator sites, e.g., if you had a POVR membership you could redirect from the public Milfvr scene to a POVR scenes.  However, if you don't have a POVR membership but you have a WankzVR one, redirecting to WankzVR  is also possible.  Since these redirects across different domains depends on what memberships they have and therefore are user specific, they have not been included.

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -486,6 +486,15 @@ func Migrate() {
 				return tx.AutoMigrate(Scene{}).Error
 			},
 		},
+		{
+			ID: "0050-members-url",
+			Migrate: func(tx *gorm.DB) error {
+				type Scene struct {
+					MemberURL string `json:"members_url" xbvrbackup:"members_url"`
+				}
+				return tx.AutoMigrate(Scene{}).Error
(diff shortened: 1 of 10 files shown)


toshski:
No migration is provided to populate the new member_url column for existing scenes. For one site they cannot be easily determined from the existing scene data.  If someone wants these for existing scenes, they can use the "Force update scenes" option to re-scrap the scenes, which will populate the column.

Issue #754, opened 2022-06-12
Easy task: hide button


mauroamore:
Easy task: add a 4th button to the scenes to hide a single scene. And a button on the left to show hidden scenes

gnadenlos:
Would be great. There are so many bonus videos from other studios or best-of-collections between regular videos, especially after the Covid situation.

toshski:
Fixed by #1093 which replaced #1060

PR #1060, opened 2022-12-13
Add is hidden status for Scenes

toshski:
This feature adds the ability to flag scenes as Hidden. A feature that has come up in the XBVR Discord a number of times.
NOTE: this PR is built on #1049 If the Property Field Filter Enhancements PR is not merged into main, I will need to rewrite parts of it and resubmit.

This feature does not attempt to hide scenes through the whole code base, it adds the ability to hide them in the Filtered Scene List/Saved Searches, where you can easily hide them (default behavior) or show them as required. This flows through to the Player APIs as they are driven by Saved Searches.

The Option needs to be enabled in Options/Web UI/Hidden Scenes

A new Property has been added to the Filters. You can filter All Scenes, Only Scenes that are not Hidden (default) and Only Hidden Scenes.

If you Disable the option, they system will revert to the current state. If you re-enable the option again, and Scenes that were previously flagged as hidden before disabling the feature, will still be flagged as hidden, they won't have to be redone.

The Quick Find and File matching lists will show all scenes; however, they will have an Icon to indicate they are flagged as hidden.

Quick Find, Hidden icon next to the Scene Title
image

File Matching Icon below the Scene Title
image


theRealKLH:
resolves #754 ?

crwxaj:
Good work, and I see the use of hiding scenes from the UI. However I have a different idea for the UI.
In my mind Hidden scenes belong to the top categories next to Any, Available, Downloaded, Not Downloaded. You could make it so the Hidden tab will only be visible, if there are actual hidden scenes in the current selection.
That way you can also scrap the option switch, since the feature shouldn't interfere if you don't use it.

toshski:
That sounds like a good ideas, I'll have a look at changing

toshski:
The UI code is probably the biggest part of this change.  It looks like it will be a lot simpler and clearer to start a new branch with the core is hidden logic changes and add the new UI changes on that, rather than remove the old UI changes and add new with this PR.
It's probably a lot less work that way as well.

toshski:
I've done the UI changes for the Hidden scene mod you suggested.  A lot less clutter on the UI and simpler.    A lot less code change as well.  I'll retest for a couple of days in day-to-day use and then release a new branch and close this one.

toshski:
Replaced by #1093.  It was clearer to go back to a clean mod based off current master than add to this PR

PR #1063, opened 2022-12-19
multifiles detection

theRealKLH:
Resolves #400 

refactored routine. db code removed.

Scenes with multiple videos and/or multiple scripts associated are detected during volume rescan

PR #1065, opened 2022-12-20
Add New Features Filtering Options

toshski:
This feature adds a range of new filter options.  A new control for Features has been added to the Filters section on the left-hand side of the Scenes.

New filterable values are similar to the Features List sent in the Heresphere Api.  It excludes some items that already have their own button in the Properties section, these also seem like options that would be used often and therefore suit having a dedicated button.

I have also added, an entry for each rating values. Ratings could be a slider, but I opted to use discrete values in the new list rather than start to clutter up the UI with more controls.  Likewise, everything is in one list, rather than needing a control for Ratings, one for FOV, one for Multiple File, one for Frame Rate, etc, again this was to avoid adding multiple controls for each feature


toshski:
This PR includes a Filter for Multiple Files which overlaps #1063

PR #1066, opened 2022-12-20
Add features filtering

toshski:
This feature adds a range of new filter options. A new control for Features has been added to the Filters section on the left-hand side of the Scenes.

New filterable values are similar to the Features List sent in the Heresphere Api. It excludes some items that already have their own button in the Properties section, these also seem like options that would be used often and therefore suit having a dedicated button.

I have also added, an entry for each rating values. Ratings could be a slider, but I opted to use discrete values in the new list rather than start to clutter up the UI with more controls. Likewise, everything is in one list, rather than needing a control for Ratings, one for FOV, one for Multiple File, one for Frame Rate, etc, again this was to avoid adding multiple controls for each feature


toshski:
Note: This PR includes a Filter for Multiple Files which overlaps https://github.com/xbapps/xbvr/pull/1063

theRealKLH:
We need a `legend` to tell what is available in "features" box... and.... do we call it "Features"? I would say these are "attributes". Is Bitrate in there?

toshski:
Happy to change the name if needed. I was actually leaning towards something like attributes, but I chose Features in the end, as there was a big overlap with the same values in the Features tags sent to Heresphere.  I probably would have chosen something else, but opted to keep the naming consistent in the absence of any obvious name. I'm not bothered either way if people think it should be different.

I'm not sure what you have in mind when you say Legend.  The list is consistent (that word again) with the other lists which just rely on the auto-complete for people to find values.  I know what you mean with "what's in the list" and I did try to get the drop down lists to be auto populated from the start so you could just press the down arrow for the autocomplete list. If you type a space and delete it, that populates the drop-down list for the down arrow, that's the effect I was trying to get automatically, ie without typing space/delete first.  But couldn't make that automatically.  I dropped it in the end and left it the same as the other files, which seems good enough for Tags, which would have a much wider range of unknown values. 

Another option which would be much more obvious is if each attribute had it's own list, eg one for Projection, one for Resolution, etc, but that would clutter up the UI with controls not often used.  Happy to look at any suggestions you've got on how to do it differently, ideally consistent with other fields. And I'll have a another try at auto-populating the drop down.

I did look at Bitrate, but when I looked into it, every video had a different bitrate.  So, without some sort of grouping, it not practical to include as you get hundreds or thousands of entries.  I don't think there is a standard for grouping for Bitrate or something everyone would agree on. This would probably be more suited to a slider type of control for filtering rather than including in a List drop down.  I considered whether Ratings would be better as a slider as well, but they have a fairly small number of entries in the list and doesn't clutter up the UI with more controls/filters than necessary. I think a separate PR for Bitrate as a Slider is probably the only practical option, someone else may want to pick that up, as I may have another unrelated project to look at

crwxaj:
I myself are also not sure about the naming. I'm not a native speaker, so calling my tag additions for HereSphere "Feature Tags" might not be the best term in the first place. Attributes or Scene properties or Video properties or whatever fits best would be fine by me.

I agree with KLH that giving some hints about what properties can be filtered there would be better, the space/delete/arrow down method works, but is pretty opaque, I'll have a look if I can find something better.

Also could you include a "no rating"/"any rating" fiter here and remove it from #1049 as I consider this filter also pretty niche-y.

toshski:
I don't think any name jumps out as it's a bit of a mixed bag.  I'll look at rename it to Attributes.

I will change the Ratings. When I did this mod, I thought I would have added it here if I hadn't already included it in #1049.  Less clutter in the UI as well.  I think maybe Trailer List as well that someone requested.  It probably wouldn't be used by enough people to warrant it's own button and could be just another Attribute filter, "In Trailer List"

toshski:
Actually, if we make Watchlist, Favorites & Scripted Attributes (as well as the existing Property buttons). We could drop #1049 altogether and leave the Property buttons as there are, rather than support Must Have/Must Not Have/Nothing.  The idea in #1049 was one consistency with the Filter buttons (Cast, Site, Tags) and more importantly to filter for things like Not Scripted. But the Attributes support the Must Have/Must Not Have so could still filter things like Not Scripted with an Attribute rather than button.  I don't think KLH is a fan of the extra toggle for Properties Buttons, so it would keep those Buttons simple but still offer the ability for the Not Have filtering

toshski:
I've managed to get the Filter List of Attributes pre-populated so a user can just press the down arrow and get a list.
We can also wrap a Tooltip around the Attribute control and list the high-level groups that a user can search.

toshski:
Note: if you added or created a Playlist while the Filter was called Features, the Playlist may not clear or set the renamed Attributes Filter

theRealKLH:
I originally had a comment here that I typed up following @crwxaj, but was waiting to commit until after testing. unfortunately my system has been running extremely slow and different containers have been fighting for memory & mysql resources that I fell asleep before I could test! I wake up to find you did much more work so I'll pull another copy and look. 

About the legend.. I was going to suggest 1. possibly just listing in README (like keyboard shortcuts), BUT who's reading the README (right)? 2. if maybe the box could be prepopulated (greyed-out style) showing example of parameter list (like one of each attribute) that would of course disappear once you begin typing.

 To me it all sounds good so far. you work on this too MUCH!

PR #1067, opened 2022-12-23
javr: replace r18.com scene importer with javdatabase.com

thebrnd:
The R18 scene importer isn't working anymore. Here's a potential replacement. Note that I don't know javdatabase.com and this parser would break easily if they change their syntax.

Usage: enter comma-separated list of one or more JAVR scene id's (like IPVR-001,SAVR-002, etc) in the "Import scene" section in the settings/options page.

### DIFF
--- a/pkg/scrape/javdatabase.go
+++ b/pkg/scrape/javdatabase.go
@@ -14,20 +14,16 @@ func ScrapeJavDB(knownScenes []string, out *[]models.ScrapedScene, queryString s
 	sceneCollector.OnHTML(`html`, func(e *colly.HTMLElement) {
 		sc := models.ScrapedScene{}
 		sc.SceneType = "VR"
-		sc.Site = "JAVR"
-		sc.HomepageURL = e.Request.URL.String()
-
-		// Cover image
-		e.ForEach(`td.moviepostermobile img`, func(id int, e *colly.HTMLElement) {
-			sc.Covers = append(sc.Covers, e.Attr(`src`))
-		})
 
 		// Cast
-		e.ForEach(`div.idol-name`, func(id int, e *colly.HTMLElement) {
-			sc.Cast = append(sc.Cast, e.Text)
+		e.ForEach(`div.idol-name`, func(id int, elem *colly.HTMLElement) {
+			sc.Cast = append(sc.Cast, elem.Text)
 		})
 
 		// Tags
+		// Always add 'javr' as a tag
+		sc.Tags = append(sc.Tags, `javr`)
+
 		// Skipping some very generic and useless tags
 		skiptags := map[string]bool{
 			"featured actress": true,
@@ -36,46 +32,67 @@ func ScrapeJavDB(knownScenes []string, out *[]models.ScrapedScene, queryString s
 			"hi-def":           true,
 		}
 
-		e.ForEach(`div.movietable tr`, func(id int, e *colly.HTMLElement) {
-			label := e.ChildText(`td.tablelabel > h3 > b`)
+		e.ForEach(`div.movietable tr`, func(id int, tr *colly.HTMLElement) {
+			label := tr.ChildText(`td.tablelabel > h3 > b`)
 
-			// Studio
 			if label == `Studio:` {
-				sc.Studio = e.ChildText(`td.tablevalue > span`)
+				// Studio
+				sc.Studio = tr.ChildText(`td.tablevalue > span`)
 
-			// Title, SceneID and SiteID all like 'IPVR-016' format
 			} else if label == `DVD ID:` {
-				sc.Title = e.ChildText(`td.tablevalue`)
-				sc.SceneID = sc.Title
-				sc.SiteID = sc.Title
+				// Title, SceneID and SiteID all like 'VRKM-821' format
+				dvdId := tr.ChildText(`td.tablevalue`)
+				sc.Title = dvdId
+				sc.SceneID = dvdId
+				sc.SiteID = dvdId
 
-			// Release date
+				// Set 'Site' to first part of the ID (e.g. `VRKM for `vrkm-821`)
+				siteParts := strings.Split(dvdId, `-`)
+				if len(siteParts) > 0 {
+					sc.Site = strings.ToUpper(siteParts[0])
+				}
+
 			} else if label == `Release Date:` {
-				dateStr := e.ChildText(`td.tablevalue`)
+				// Release date
+				dateStr := tr.ChildText(`td.tablevalue`)
 				tmpDate, _ := goment.New(strings.TrimSpace(dateStr), "YYYY-MM-DD")
 				sc.Released = tmpDate.Format("YYYY-MM-DD")
 
-			// Tags
 			} else if label == `Genre(s):` {
-				e.ForEach(`td.tablevalue > span.tags`, func(id2 int, e2 *colly.HTMLElement) {
-					tag := strings.ToLower(e2.Text)
+				// Tags
+				/* NOTE:
+				   "Tags are technically incomplete vs. what you'd get translating dmm.co.jp
+				   tags/correlating them back to their old equivalents on r18 using something
+				   like Javinizer's tag CSV"
+				*/
+				tr.ForEach(`td.tablevalue > span.tags`, func(id2 int, span *colly.HTMLElement) {
+					tag := strings.ToLower(span.Text)
 
 					if !skiptags[tag] {
 						sc.Tags = append(sc.Tags, tag)
 					}
 				})
 
-			// Synopsis / description
 			} else if label == `Translated Title:` {
-				sc.Synopsis = e.ChildText(`td.tablevalue`)
+				// Synopsis / description
+				sc.Synopsis = tr.ChildText(`td.tablevalue`)
+
+			} else if label == `Content ID:` {
+				contentId := tr.ChildText(`td.tablevalue`)
+				sc.HomepageURL = `https://www.dmm.co.jp/digital/videoa/-/detail/=/cid=` + contentId + `/`
+				sc.Covers = append(sc.Covers, `https://pics.dmm.co.jp/digital/video/`+contentId+`/`+contentId+`pl.jpg`)
 			}
 		})
 
 		// Screenshots
-		e.ForEach(`img`, func(id int, e *colly.HTMLElement) {
-			alt := e.Attr(`alt`)
-			if strings.Contains(alt, "Screenshot") {
-				sc.Gallery = append(sc.Gallery, e.Attr(`src`))
+		e.ForEach("a", func(_ int, anchor *colly.HTMLElement) {
+			linkHref := anchor.Attr(`href`)
+			/* NOTE:
+			   it only pulls 6 gallery images, but that appears to be a limitation
+			   of how javdatabase.com is set up, they only pull 6 gallery images.
+			*/
+			if strings.HasPrefix(linkHref, "https://pics.dmm.co.jp/digital/video/") && strings.HasSuffix(linkHref, `.jpg`) {
+				sc.Gallery = append(sc.Gallery, linkHref)
 			}
 		})
 
--- a/pkg/scrape/javdatabase.go
+++ b/pkg/scrape/javdatabase.go
@@ -1,0 +1,92 @@
+package scrape
+
+import (
+	"strings"
+
+	"github.com/gocolly/colly"
+	"github.com/nleeper/goment"
+	"github.com/xbapps/xbvr/pkg/models"
+)
+
+func ScrapeJavDB(knownScenes []string, out *[]models.ScrapedScene, queryString string) {
+	sceneCollector := createCollector("www.javdatabase.com")
+
+	sceneCollector.OnHTML(`html`, func(e *colly.HTMLElement) {
+		sc := models.ScrapedScene{}
+		sc.SceneType = "VR"
+		sc.Site = "JAVR"
+		sc.HomepageURL = e.Request.URL.String()
+
+		// Cover image
+		e.ForEach(`td.moviepostermobile img`, func(id int, e *colly.HTMLElement) {
+			sc.Covers = append(sc.Covers, e.Attr(`src`))
+		})
+
+		// Cast
+		e.ForEach(`div.idol-name`, func(id int, e *colly.HTMLElement) {
+			sc.Cast = append(sc.Cast, e.Text)
+		})
+
+		// Tags
+		// Skipping some very generic and useless tags
+		skiptags := map[string]bool{
+			"featured actress": true,
+			"vr exclusive":     true,
+			"high-quality vr":  true,
+			"hi-def":           true,
+		}
+
+		e.ForEach(`div.movietable tr`, func(id int, e *colly.HTMLElement) {
+			label := e.ChildText(`td.tablelabel > h3 > b`)
+
+			// Studio
+			if label == `Studio:` {
+				sc.Studio = e.ChildText(`td.tablevalue > span`)
+
+			// Title, SceneID and SiteID all like 'IPVR-016' format
+			} else if label == `DVD ID:` {
+				sc.Title = e.ChildText(`td.tablevalue`)
+				sc.SceneID = sc.Title
+				sc.SiteID = sc.Title
+
+			// Release date
+			} else if label == `Release Date:` {
+				dateStr := e.ChildText(`td.tablevalue`)
+				tmpDate, _ := goment.New(strings.TrimSpace(dateStr), "YYYY-MM-DD")
+				sc.Released = tmpDate.Format("YYYY-MM-DD")
+
+			// Tags
+			} else if label == `Genre(s):` {
+				e.ForEach(`td.tablevalue > span.tags`, func(id2 int, e2 *colly.HTMLElement) {
+					tag := strings.ToLower(e2.Text)
+
+					if !skiptags[tag] {
+						sc.Tags = append(sc.Tags, tag)
+					}
+				})
+
+			// Synopsis / description
+			} else if label == `Translated Title:` {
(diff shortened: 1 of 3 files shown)


vt-idiot:
I tried this out in Gitpod after a bunch of hair pulling. Please let me preface my nitpicking by saying "amazing!" - it really works great. A few nitpicking observations:

1. Scene title is just the JAV ID. That's OK, but it does put the actual title into the description field, meaning it does have it in there somewhere.
2. Scene links to JavDB. Would be nice to link directly to dmm.co.jp, but it's hardly worth worrying about.
3. Cover image is set to a webp image hosted at javdatabase.com - **this is bad** and _should be changed_
4. Site/studio behavior breaks norms with the old JAVR R18 scraper and should be changed IMO
5. Tags are technically incomplete vs. what you'd get translating dmm.co.jp tags/correlating them back to their old equivalents on r18 using something like Javinizer's tag CSV, but "eh", the pros far far outweigh the cons here.

Regarding 4:

- the "Site" is set to JAVR, but the studio is set correctly.
  - This will lump all JAV together - I know "VRKM" - I don't think to filter by "K M Produce" - nor is there any way to filter by "studio" in XBVR anyways.
- The old scraper set the "prefix" as the site.
  - So scraping e.g. VRKM-821 would set the site to "VRKM".
  - In all caps, because setting it to "vrkm" would create a separate entry in the database...
- "javr" used to be added as a tag.

The actual cover image can be derived from the gallery images, since javdatabase.com doesn't actually host those, and links directly to dmm.co.jp anyways. e.g. again for JAV ID VRKM-821, this scraper gives gallery images like `https://pics.dmm.co.jp/digital/video/vrkm00821/vrkm00821-1.jpg` - the cover image is then just  `https://pics.dmm.co.jp/digital/video/vrkm00821/vrkm00821pl.jpg`. The scene URL can also be similarly derived `https://www.dmm.co.jp/digital/videoa/-/detail/=/cid=vrkm00821/` - this is also always true.
This is _always_ true, a hypothetical "ABCDE-12345" with gallery images like `https://pics.dmm.co.jp/digital/video/abcde12345/abcde12345-1.jpg` will have a cover image of `https://pics.dmm.co.jp/digital/video/abcde12345/abcde12345pl.jpg`.

vt-idiot:
also a minor thing - it only pulls 6 gallery images, but that appears to be a limitation of how javdatabase.com is set up, they only pull 6 gallery images.

vt-idiot:
...and the image links it's serving seem to be thumbnails:
`https://pics.dmm.co.jp/digital/video/vrkm00821/vrkm00821-1.jpg` gallery image for ants
`https://pics.dmm.co.jp/digital/video/vrkm00821/vrkm00821jp-1.jpg` proper image

thebrnd:
Thank you for taking a look! I patched issues 2, 3, 4 and the unnumbered low-res thumbnail issue.

I left the title as 'VRKM-821' style with the translated long version in the description for now as I wasn't sure how to interpret issue number 1. Personally I prefer having the short title the way it is here but it's a matter of preference - as you said it's easy to change if needed. I also did not address the "incomplete" tags and "incomplete" gallery issues for now, but added comments into the code as those are potential future improvements.

vt-idiot:
I've been importing JAV scenes manually since June, and SOD's scenes for about a year so anything would've been an improvement, but with those fixes in place, now I can't wait for it to go live. Great work. The tags and 6 image galleries aren't huge issues either, the scraper is doing what it should, that's on JAVDatabase to change.

This became a huge TL;DR, so I'll summarize it with my last request, which is hopefully a quick and easy fix: add a "sanity check" of sorts for one studio. This is also technically JAVDatabase's fault.

### Some SOD Create/**_3DSVR_-** scenes are listed "oddly" at JAVDatabase.com.
Here are two scenes as examples, neither of their scene pages should be blocked while accessing dmm.co.jp from outside of Japan: using the "full" ID as seen in a DMM image URL, say you have `13dsvr01192` from SOD and `dsvr00037` from Das.

- There's no ambiguity with the Das scene
  - Searching DSVR-037 on javdatabase.com takes you directly to the correct scene
  - Trying to scrape DSVR-037 using the new scraper works fine, Studio = Das, Site = DSVR
- The SOD scene throws a wrench into things
  - javdatabase has it as `DSVR-01192`, lists the DVD ID as `DSVR-01192` as well
  - does show the full content ID - `13dsvr01192`
  - it scrapes into XBVR with a scene ID of `DSVR-01192` and `DSVR` as the site - this will group it with Das's scenes
  - JAVR via R18 would've scraped it into XBVR as `3DSVR-1192`, `3DSVR` as the site
## proposal? idk if it's feasible
- _If_ the "content ID" in the DMM image URL _after_ a successful scrape matches `13dsvr0\d{4}` - so 13dsvr00001 -> 13dsvr09999
- _And_ the scraper thinks the site ought to = `DSVR`
- only if **both** are true, _then_
  1. Studio should be overwritten (if it's necessary, I forgot) = SOD Create
  2. Site should be overwritten - this one is necessary = 3DSVR
  3. Scene ID should - this is ultimately cosmetic = 3DSVR-\d{4}

If you want a quick set of test IDs that should all scrape successfully, not necessarily "correctly", `dsvr-01192` `3dsvr-1000` `dsvr-037` and `3dsvr-0878` should be enough to illustrate the issue. Now, as for the why, here's the mucho texto...

<details>
  <summary>mucho texto</summary>

Entry iii. on the "fix" is purely cosmetic, because Das is using 3 digit numbering for their `DSVR` scenes, and as far as I can tell, javdatabase only returns DSVR-01XXX for SOD Create scenes:
- above 3DSVR-1000 or thereabouts
- always with five digits

So there will _never_ be a collision, and it really is only cosmetic. And I guess it affects search results when file matching too. Studio might actually already populate properly, but the site incorrectly populates as DSVR, grouping SOD Create's newer scenes and Das's scenes together as a result.

## additional irrelevant examples  
  - R18 would've had it as `3DSVR-1192` but would've shown it anyways if you searched `13dsvr01192`
  - Afesta (example) returns the correct scene when searching `3DSVR-1192`
  - JAVBUS (another example) also does, `https://www.javbus.com/en/3DSVR-1192`
  - I couldn't get dmm's japanese search box to show me the scene until I actually searched for `13dsvr01192`, lol
  - JAVLibrary (bad example) has always done it _wrong_, but luckily their Cloudflare shenanigans would've prevented it from being used as a good XBVR scraper anyways
    - they don't even have this scene yet...
    - ex. 1 3DSVR-0878 is in here as `DSVR-878`
    - ex. 2 search `DSVR-037`
    - there are two of them: `https://www.javlibrary.com/en/?v=javmefyy4i` for the actual Das DSVR-037 and `https://www.javlibrary.com/en/?v=javlikql5m` for what should be labeled 3DSVR-0037
    - both show as `DSVR-037` - so a hypothetical JAVLibrary scraper would've overwritten one scene for the other

## thinking out loud to try and figure out how/why it works...
> Maybe they always did it this way at javdatabase?

*nope.avi* - `https://www.javdatabase.com/movies/3dsvr-0878/` - note: adding the leading zero here is a non-issue, you would've had to add it just to get it to show up in the search box on R18, and that is how JAVR scraped that particular scene anyways

> Maybe they switched once they rolled over 4 digits

*nope.avi* - `https://www.javdatabase.com/movies/3dsvr-1000/`
> Maybe they switched over at 1100?

*nope.avi* - `https://www.javdatabase.com/movies/3dsvr-1099/` works, `https://www.javdatabase.com/movies/3dsvr-1101/` works. 3DSVR-1100 isn't actually on here at all, nor is it on here as DSVR-01100. The scene does exist though.

> Maybe it's got something to do with geo-restricted scenes?

*nope.avi* - 3DSVR-1100 and 3DSVR-1166 are both geo-restricted, the former is missing on javdatabase, the latter is present
`https://www.dmm.co.jp/digital/videoa/-/detail/=/cid=13dsvr01100/` lists a release date about a month late, and shows a geo-restriction error if you're not using a Japanese proxy/VPN. Might be exclusive to either Afesta or SLR outside of Japan. Browsing JAV on SLR is a pain in the ass, so I'm not about to try and find it, but thar she blows: `https://www.afesta.tv/en_vr/face.php?fid=077a99bb1239b2182706e701887158bd`
`https://pics.dmm.co.jp/digital/video/13dsvr01100/13dsvr01100jp-1.jpg` luckily the images will load fine even in a geo-restricted location. ~~If I were a betting man, I'd bet that their scraper is in the west, so geo-restricted scenes are probably all missing on JAVDatabase.com~~ nope, their URLs are just case sensitive...

- `https://www.dmm.co.jp/digital/videoa/-/detail/=/cid=13dsvr01166` - accessible in Japan
- `https://www.javdatabase.com/?s=dsvr-01166` - **works**
- `https://www.dmm.co.jp/digital/videoa/-/detail/=/cid=13dsvr01166` - without a VPN:
> image
> このサービスはお住まいの地域からは
> ご利用になれません。
> Sorry! This content is not available in your region.

## probably can't fix this
I was initially trying to think of a creative solution for re-writing search box entries that would always result in a 3DSVR scene properly being scraped, no matter what goes into the search box. As for knowing what to actually type into the box to get it to scrape a SOD Create VR scene, if the code is <900 it should be easy. `3DSVR-0123` will work like it did before, and it's true all the way down to the very first one - `https://www.javdatabase.com/movies/3dsvr-0001/`. Above that, the gaps start, it gets confusing, and impossible to solve, because if there is rhyme or reason to this on javdatabase's part, I haven't found it yet.

I tried checking to see where the old one "stops", and since they used a "Label" of SODVR on javdatabase.com, you can easily see that: `https://www.javdatabase.com/labels/sodvr/`
- The 3DSVR-XXXX scenes stopped on May 31st - the last day R18 added new scenes...
- The newer DSVR-0XXXX entries are missing their "Label" field entirely.
- `https://www.javdatabase.com/movies/dsvr-01129/` DSVR-01129 exists on javdatabase, but `https://www.javdatabase.com/movies/3dsvr-1135/` 3DSVR-1135 and `https://www.javdatabase.com/movies/3dsvr-1127/` 3DSVR-1127 do too : there's no easy "sequential check" that could be done
- Some scenes like 3DSVR-1026 and 3DSVR-1100 mentioned earlier are missing entirely anyways.

It's worth mentioning that both of those scenes never made it onto R18 for whatever reason either, and 1026 just now made it onto DMM with a reported release date of 2022/11/24; the scene is from 2021/09/27. Something tells me those "gap scenes" are probably all missing from javdatabase and probably wouldn't get added in. I don't feel like checking them all, I already wrote JSON manifests for the ones in my library, but the other scenes around it missing from the big R18 dump I grabbed are also missing from JAVDatabase - no 3DSVR-1025, no 3DSVR-1028. 1027 is missing too, and R18 _did_ eventually add that one.

## New proposal
* If scraper input matches `*DSVR*`
* Print "**Data → 🖕(⸌益⸍)🖕**" in the status bar
* Crash xbvr
</details>

IIRC somewhere around the 900's, SOD Create's VR scenes started getting added to R18 very erratically, so I'm not surprised it's true on DMM too. Maybe there's an ongoing tiff between DMM and SOD over $$$, maybe it's exclusivity deals or something, but **even when they don't have a scene publicly listed on dmm.co.jp in Japan, the image URLs are always live...**

OK, so users will just have to check and see manually on javdatabase for ~~SODVR~~ _can't call them that either_, because they used it as a code in the past...`SOD Create's VR scenes` to see what's up if they have an ID >~900 or thereabouts. No big deal.

Besides, you can always just wait for SOD Create to give SLR the go ahead to bring the scene to the west 12-24 months later and scrape it that way! /s

vt-idiot (changes requested):
Here's one more incredibly annoying, generic and useless tag to have it drop (great idea BTW): `exclusive distribution`

From a cursory glance it looks like they already drop the even more annoying ones like `k m produce / alice japan 30% off sale` and `mousouzoku / celeb no tomo 30% off sale`.

pkg/scrape/javdatabase.go
```diff
@@ -0,0 +1,109 @@
...
+		// Tags
+		// Always add 'javr' as a tag
+		sc.Tags = append(sc.Tags, `javr`)
+
+		// Skipping some very generic and useless tags
+		skiptags := map[string]bool{
+			"featured actress": true,
+			"vr exclusive":     true,
+			"high-quality vr":  true,
+			"hi-def":           true,
```
vt-idiot, line 32:
```go
			"exclusive distribution":           true,
```

thebrnd:
Patched the 3DSVR categorization issue and the extra tag filter. The studio appears to already be set to "SOD Create" so I left out overwriting that part.

Regarding the incomplete tags and gallery count: I was planning to perhaps use javlibrary.com instead to solve that, but after reading the mucho texto part I gather that this might not be a great idea. The javlibrary parser is already implemented though (in the javlibrary branch), and I didn't encounter any cloudflare issues (it just works here), but I suppose those might show up at some point.

The missing scenes in javdatabase.com is the biggest issue for me. Those 3DSVR scenes are not the only gaps in their database. If you have one of those scenes and you also use xbvr it's a problem. Having both javdatabase.com plus javlibrary.com seems to 'solve' this for the most part, although there are still scenes that are missing from both. In the javlibrary branch I actually added both scrapers and give the xbvr-end-user the choice of which one to use for importing. The tags and cast-names are different, so you do get inconsistent data when mixing these two. I'm open to ideas on what to do with the javlibrary branch, but I'll leave it out of this pull request for now to keep things simple.

vt-idiot:
Oh, you already implemented one for JAVLibrary, nice. I never have any issues browsing the site itself, but when Javinizer tries to scrape it, I run into issues and have to fiddle with Cloudflare cookies to make it work again. Unfortunately I got nothing.

The only other mostly complete, definitely more than javdatabase, 3rd party JAV "db" I'm aware of is javbus.com.
1. They always have links to the full DMM gallery
2. They host the cover image themselves, but you already have resolving the cover URL figured out
3. They seem to use tags that largely match what R18 used to use, e.g. I see `寝取り・寝取られ・NTR` from 3DSVR-1166 mapped to "cheating wife"
4. **Their actress names are all over the place** - the only downside

Speaking of javinizer, if you're not familiar with it, and you understand PowerShell, it might be worth looking at for ideas instead of re-inventing the wheel - [jvlflame/Javinizer](https://github.com/jvlflame/Javinizer) 
It has the option to scrape multiple sites and combine the metadata, and it's how I was making "XBVR JSON manifests by hand" go a lot faster than actually doing them by hand. It does require user input though, since it's more of a full fledged metadata tool. Most of the other metadata sources it scrapes are pretty useless for VR JAV, but they do use, for example, a .csv to "match" actress names, to what R18 would've called them. 

Anyways, don't fret about making it "perfect" - there will always be some element of scraping this stuff "by hand" for more than one reason. HUNVR and all of the associated "labels" love to put scenes out with **no cast information**, so I have to add them in, and if I don't recognize the actress, go to e.g. sougouwiki and figure out who's in it myself.

vt-idiot (approved):
nice, all bases covered

crwxaj:
@thebrnd @vt-idiot Would you consider this PR ready to be merged?

vt-idiot:
Unless @thebrnd was planning on also including the JAVLibrary branch scraper in this PR, it seems great to me, and the minor issues I found in testing were all addressed.

javdatabase has some gaps but that's nothing the scraper itself can fix without using a different source entirely, and JAVLibrary has its own plethora of issues, which is why I imagine this was the one in the PR.

crwxaj:
I gave it a try, and I'm afraid the scraper doesn't work anymore. Did they change their site already?
For example `label := tr.ChildText("td.tablelabel > h3 > b")` doesn't match anymore, because there is no <h3> Tag in the the table. The cast was also missing.

@thebrnd please have another look

thebrnd:
Yes, they seem to have changed their website. It looks different visually now than before - we were just lucky it happened before the merge instead of right after. Hopefully they don't do that too often.

The scraper is updated now, again, so it looks like we need another round of reviewing.

If anybody wants context on the javdatabase vs javlibrary issue here is the mucho texto part: it wasn't really a choice on my side preferring one over the other. The XBVR 0.3.31 upgrade broke for me, it wouldn't start and not even print a single line in the docker logs, when using the existing database. I had to clear the database and start fresh for XBVR 0.3.31 to work. I figured rebuilding the database was less effort than debugging whatever the issue was, but forgot that it wouldn't be able to import javr scenes anymore... At this point, implementing the scraper seemed like the easiest solution for it. Since I didn't know of any good sources I found javdatabase.com at random and implemented something quickly. After doing so, vt-idiot pointed out some good feedback that couldn't be solved with the javdatabase.com, so I found another random site (javlibrary) and implemented a parser for that also, hoping to solve all issues. But.. neither of these two sources is complete, and they use different tags, so mixing both poses some "issues". Then I chose to postpone working on that part until after this merge request, as I couldn't know what further feedback was going to happen and it might be relevant for the other importer as well. If this importer ever gets merged I might have more pull requests at some point. Maybe just a dropdown selection box of which javr importer you want to use so there is some fallback option when one goes down, changes and/or is missing a scene.

Regarding the question about whether I consider this ready to be merged: I would have answered that I consider the data-centric review handled, plus the black box testing. On code level, I have never used golang before this and didn't spend much time studying several assumptions that I made during development. For example, I didn't really check how XBVR and golang deal with threading, and I added sceneCollector.Wait() to delay returning from the importer function until all scenes have been scraped. I assumed that this wouldn't block the whole xbvr process but only the scraping thread, but since scraping happens so fast I never actually verified this. I figured, somebody who has worked on golang and XBVR could code-review this within 2 minutes. I am also putting some trust in XBVR's ability to leave out scenes that weren't fully scraped as this was working during testing, but I am not familiar with code policies you might have on that and what you expect from scrapers in detail. If nobody wants to do any code reviews, I'd still merge this anyway, as it's more useful than the current r18 importer.

PR #1072, opened 2023-01-01
fix vrporn not finding scenes

theRealKLH:
VRPorn not finding any scenes due to html code change.


theRealKLH:
did not see #1057

PR #1073, opened 2023-01-01
Clear All Process Locks at Startup

toshski:
When XBVR starts existing Process Locks are removed.  Process locks for "files" and "heatmaps" were not included in the list of locks removed
This was not an issue before, because the issue addressed in #1061 where processes tried to run and but was still locked would still have their lock removed and allowed it to run next time.

I have changed the startup code in server.go to remove kvs starting with 'lock-' rather than just add to the list of locks to remove.  This reduces the chance of new Process Locks been added in the future and missing they need to removed at startup.

### DIFF
--- a/pkg/models/db.go
+++ b/pkg/models/db.go
@@ -1,6 +1,8 @@
 package models
 
 import (
+	"strings"
+
 	"github.com/avast/retry-go/v3"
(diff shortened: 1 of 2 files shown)


toshski:
Need to rebase this off 1061

PR #1093, opened 2023-01-05
Is hidden (using dl state)

toshski:
This feature adds the ability to flag scenes as Hidden. A feature that has come up in the XBVR Discord a number of times.
This feature does not attempt to hide scenes through the whole code base, it uses a new Download State tab to separate them out from unhidden scenes.

A scene can be flagged as hidden in the Scene List or Scene Detail. 
image

The Quick Find and File matching lists will show all scenes; however, they will have an Icon to indicate they are flagged as hidden.

Quick Find, Hidden icon next to the Scene Title
image

File Matching Icon below the Scene Title
image

If a scene is flagged as Hidden, it will only be listed under Hidden, e.g. a Hidden scene with a downloaded video, it will not be listed under Downloaded.  I think this best reflects the intention people would usually have when hiding a scene.  

This PR replaces #1060

### DIFF
--- a/pkg/api/heresphere.go
+++ b/pkg/api/heresphere.go
@@ -412,6 +412,10 @@ func (i HeresphereResource) getHeresphereScene(req *restful.Request, resp *restf
 		addFeatureTag("Has cuepoints")
 	}
 
+	if scene.IsHidden {
+		addFeatureTag("Hidden")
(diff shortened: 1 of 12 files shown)


vt-idiot:
Fixes #754

PR #1095, opened 2023-01-06
Site Subscriptions

toshski:
This feature allows users to flag Studios for which they have a subscription.  This is done with a new Toggle switch under Options/Scrapers.

Once enabled any scenes for that scraper are highlighted with a green border to indicate they have a subscription. e.g. screenshot of a scene with and without subscription.  
image

There may be a better to highlight the scenes with a subscription (I think it was my first mod) maybe an icon in the top/left or right.  I did consider adding an extra icon underneath like Favorite/Watched, but it just blends in, and you don't notice it. I keep thinking about changing it, but it seems to work without distracting or cluttering up the UI. UI is probably not my strength so open to any other options.

I have also added a button to toggle the Subscription status of all visible Studios.  This is mainly for the SLR Studios as it takes ages to Toggle each studio individually.  Users can type SLR in the Source and then Toggle all subscriptions. a lot easier.  On the negative side I think it looks slightly out of place and clutters things a bit.  It could be dropped as it is just a convivence thing and doesn't affect the functionality.

To make this work, I needed to establish a link between the site and scene.  You could link this 99% off the time, but there was a  couple of small sites where this couldn't be done, e.g. Little Caprice Dreams.  A migration will update the scraper id on each scene.  There is a progress message as each studio is migrated, but it actually pretty quick anyway.

Note: I've tried to guess where numbering will be at for the migrations, but that may need changing when finally merged with master


toshski:
If Features Filtering mod goes in, I'll add 2 Feature for "Has Subscription" and "No Subscription"

toshski:
Assuming #1113 goes in before this PR. I will need to recheck this PR.  Some code (including the migration) from this PR was used in #1113 so there may be conflicts and the migration will need updating.

toshski:
This has now been updated with changes needed due to some of the migration code for the scraper_id been incorporated into PR #1113

crwxaj:
I'm also not an UI designer, but I found the bordered boxes a bit too disruptive in the scene list. I wanted something smaller, but still noticeable, so I highlighted the studio name. What do you think?

toshski:
I like that and you said you weren't a UI designer

PR #1102, opened 2023-01-07
search index: faster method of adding newly scraped scenes

thebrnd:
After adding 1 or more scraped scenes from the JAVR scene importers, XBVR would check all existing scenes in the database for their presence in the search index. If you have scraped a large amount of scenes, this takes a very long time after every single scene import.

This commit changes that by _only_ adding the newly scraped scenes into the index.

Furthermore, the previous code would keep the index as-is if a scene with the same SceneID already existed in the index. But if you re-scrape the same scene, the data may have been updated and the search index should now be updated as well.

This commit only uses the new code for the JAVR scene importers to start with, but in theory the other importers can use the same new method after a scrape as well..

### DIFF
--- a/pkg/api/scenes.go
+++ b/pkg/api/scenes.go
@@ -181,7 +181,7 @@ func (i SceneResource) createCustomScene(req *restful.Request, resp *restful.Res
 		return
 	}
 
-    // Update search index with new scene
+	// Update search index with new scene
 	scenes := []models.Scene{resultingScene}
 	tasks.IndexScenes(&scenes)
 
--- a/pkg/api/scenes.go
+++ b/pkg/api/scenes.go
@@ -723,6 +723,10 @@ func (i SceneResource) editScene(req *restful.Request, resp *restful.Response) {
 
 		scene.Save()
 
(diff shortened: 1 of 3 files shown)


vt-idiot:
One of the things I noticed was that after the change from the "old" import some externally hosted `.json` manifest to the new import/export function, was that before, I could import a dozen .json links in a row, scrape one JAVR scene, or force an update on any of the other scrapers to force the re-index, and be done.

With the new JSON importer, even if you import a manifest with one scene in it, you have to wait for, oh I don't know, 20k scenes to be re-indexed, every single time. This is great. Would there be a way to expand the functionality to the manifest importer? Plus now I guess I have a second thing to play around with tomorrow.

tosh6072:
Another gap in the Search Index logic I've been meaning to look at some time, is a user can edit a scene.  May be worth adding a call to your function from the editScene function in pkg/api/scenes.go

thebrnd:
The scenes imported from the JSON updater are now being indexed using the new method.

Additionally, after editing a scene manually, the index is also updated.

PR #1107, opened 2023-01-19
JAVR Tag Update

vt-idiot:
For the sake of avoiding clutter/multiple PRs, would it be possible to leave this one open until right before the next release? I might make a few more commits over the next week or so as I scrape the last few unmatched JAVR scenes in my library and encounter more tags. Also thought about opening a tracking ticket for this but wasn't sure if it was worth it.

TL;DR The new JAVR scrapers both drop some annoying tags at scrape time, and make some tag substitutions.

1. Added one new entry, `dmm exclusive`, to the drop list.
2. Added two new tag substitutions for cases where JAVLibrary uses a different tag than JAVBUS and JAVDatabase, but most importantly, different vs. the old R18 tags.

I have ~1-2k old scenes that were scraped from R18, as well as a JSON dump (unfortunately not as XBVR manifests) of all of R18's VR scene data from right before it closed, so that's the source I'll be focused on for tag substitution. "If the tag existed on R18, that one is automatically the 'right' one.

### DIFF
--- a/pkg/scrape/javutil.go
+++ b/pkg/scrape/javutil.go
@@ -38,15 +38,16 @@ func ProcessJavrTag(tag string) string {
 	// Map some tags to normalize so different sources match
 	// TODO: this mapping is totally incomplete and needs help from community to fill
 	maptags := map[string]string{
-		"blow":         "blowjob",
-		"blow job":     "blowjob",
-		"kiss":         "kiss kiss",
-		"kiss / kiss":  "kiss kiss",
-		"prostitute":   "club hostess & sex worker",
-		"prostitutes":  "club hostess & sex worker",
-		"sun tan":      "suntan",
-		"huge cock":    "huge dick - large dick",
-		"other fetish": "other fetishes",
+		"blow":               "blowjob",
+		"blow job":           "blowjob",
+		"kiss":               "kiss kiss",
+		"kiss / kiss":        "kiss kiss",
+		"prostitute":         "club hostess & sex worker",
+		"prostitutes":        "club hostess & sex worker",
+		"sun tan":            "suntan",
+		"huge cock":          "huge dick - large dick",
+		"other fetish":       "other fetishes",
+		"threesome/foursome": "threesome / foursome",
 	}
 	if maptags[taglower] != "" {
 		return maptags[taglower]
--- a/pkg/scrape/javutil.go
+++ b/pkg/scrape/javutil.go
@@ -44,6 +44,7 @@ func ProcessJavrTag(tag string) string {
 		"prostitute":  "club hostess & sex worker",
 		"prostitutes": "club hostess & sex worker",
 		"sun tan":     "suntan",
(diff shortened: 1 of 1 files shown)


vt-idiot:
On a somewhat related note, @thebrnd, would a system similar to the tag substitutions be possible for _studio_ substitutions in general, or even actresses in the long run? Perfect example for studios, since there are at least three unique results:

### CLVR e.g. `https://www.javbus.com/en/CLVR-041`
| Source | Result |
| --- | --- |
| JAVBUS | `HentaiShinshiClub` |
| JAVDatabase | `Hentai Shinshi Club` |
| jav.land | `Hentai Gentleman Club` |
| JAVLibrary | Down for maintenance... |
| **R18** | **[`Hentai Shinshi Club`](https://github.com/xbapps/xbvr/files/10459806/clvr00041.json.txt)** |

JAVLibrary was either `HentaiShinshiClub` or `Hentai Shinshi Club`, I forgot. Point being that if it existed on R18, that should take priority?

so something like:

```go
	mapstudios := map[string]string{
		"Hentai Gentleman Club":  "Hentai Shinshi Club",
		"HentaiShinshiClub":      "Hentai Shinshi Club",
	}
```

thebrnd:
> On a somewhat related note, @thebrnd, would a system similar to the tag substitutions be possible for studio substitutions in general, or even actresses in the long run?

In general I would say the the code is not the right place to include data such as this. We made an exception for the tags to keep things simple, and hoping that this mapping doesn't need to change much. But adding studios and actresses will create an ever-growing list and would burden the XBVR maintainers with many pull requests they may grow tired of.

So, while I think the feature of having such mappings is nice to have, I think it would require a different kind of implementation. Ideally one where updates to this data can be published and maintained without updating the source code.. and that would require some form of hosting of this dataset, which is a complication and I don't immediately know how to solve that.

vt-idiot:
Fair enough.

vt-idiot:
> > So, while I think the feature of having such mappings is nice to have, I think it would require a different kind of implementation.
> 
> Isn't this the sort of thing AKA groups are for? Maybe those could be done for studios too. (But I've never used that feature myself, so it might not work the way I think it does.)

Having AKAs for **tags** would eliminate the need for anyone to update the tag substitution rules in javutil.go, beyond making sure it continues to drop the useless ones like `dmm exclusive`. Studio wouldn't be worth implementing AKAs for really, or even editing them if JAVR scrapes them in Japanese or with/without spaces or something. The filter in xbvr's UI is for **Site**, `Studio` isn't really visible unless you're editing a scene, and I don't think it's actually used for anything unless you're directly accessing main.db and doing an SQL query or filtering in there or something.

vt-idiot:
thanks @crwxaj

when I get ~~a minute~~ an entire ~~afternoon~~ day to mess with this, I'll open a follow-up, far more thorough PR that scours a few dozen saved R18 json's vs. the new JAVR scrapers we currently have implemented and aligns the tags they all use

PR #1112, opened 2023-01-26
VRHush/VRAllure updates to Cover/Title code

theRealKLH:
Fix cover images.

Latest videos (vrhush-0434 & vrhush-0435) are missing 960x540 sized cover image. Change to use the default 1110x624 size.


alexnsfw:
The same change is needed in the VRAllure scraper too.

theRealKLH:
> The same change is needed in the VRAllure scraper too.

good catch!

PR #1113, opened 2023-01-29
Add configurable Scraper list

toshski:
This feature allows users to update the list of Studios scraped from the aggregator sites, SLR, VRPorn, POVR & VRPHub
This will allow users to 

- add a studio without needing to wait for a new release
- add a studio for a minor/niche studio that may never be added to the official release
- use an alternate aggregator site for a studio

This list of studios has been shifted from been hard coded, to a Json configuration file, scrapers.json.  The file is embedded in the xbvr executable and will be saved to the users App directory.  The file has 2 sections, one for the xbvr official list of studios and a second for a custom list the user can maintain.  If the user changes the official list, it will be overwritten with the proper list.

If a studio in the user's custom list is released in the official list, then the entry in the custom list will be removed, if the studio was for the same aggregation site, or to have no scenes. If the user had specified a different Name or Company, the existing scenes will be flagged to be re-craped, so the scenes are updated with the correct values.  If the user used a different aggregation site, they would need to deleting existing scenes to change to the official list, this is because scene id will be different, and you cannot 100% match scenes between two different aggregators.  If they don't delete the existing scenes, they will continue to scrap from the alternate aggregation site.

Users should not share scene data from a studio in a custom list.  If the studio is eventually released and from a different site, the scene ids will be different. An option has been added to the Import/Export function to only include Studios from the official list if sharing with someone else.  The default is still to include all studios.

If a user scans some scenes for a studio in their custom list and then wishes to change the Studio Name or Company, they need to first flag the existing scenes to be re-scraped, update the studio's name/company, restart xbvr, then run the scraper. This will update the scenes with the new name/company.  If they forget, they can change the studio fields back in the scraper.json file, then follow the correct procedure.

To add a studio to the official list, update the scraper.json file in the pkg/config directory.  For a user to add a studio to their custom list they can edit the scraper.json file in their xbvr directory.

Warnings, re: changing the official studio list, not sharing scenes, renaming the Name/Company and the possibility of needing to delete scenes and rescan if the official list uses a different aggregator, have been included at the top of the scraper.json file.


pkg/config/scraper_list.go
```diff
@@ -0,0 +1,159 @@
...
+	CusomtScrapers CusomtScrapers `json:"custom"`
+	XbvrScrapers   XbvrScrapers   `json:"xbvr"`
+}
+type XbvrScrapers struct {
+	PovrScrapers   []ScraperConfig `json:"povr"`
+	SlrScrapers    []ScraperConfig `json:"slr"`
+	VrpornScrapers []ScraperConfig `json:"vrporn"`
+	VrphubScrapers []ScraperConfig `json:"vrphub"`
+}
+type CusomtScrapers struct {
```
crwxaj, line 29: There is a typo

crwxaj:
I really like this PR. It would give users the ability to quickly add some SLR scrapers on their own, and we might even consider to remove some of the more obscure and abandoned studios from the official scraper list.

Regarding scene IDs: I don't like studio specific scene_id prefixes, since they don't add value, and raise some issue, as you pointed out above. I think this might be a good time to get rid of them and move to scraper specific IDs. For example "slr-originals-31862" would become simply "slr-31862" and "lustreality-31341" would become "slr-31341". This would make these IDs globally unique, even for custom scrapers. The main issue would be that previous data dumps wouldn't match anymore, but I think it's the better choice going forward.

What do you think?

toshski:
My main concern is people having different Ids for the same scene, just because they scraped from a different aggregator. e.g. "Latina Girl Will Make Your Day", is lustreality-24488 and would become slr-24488, but if someone had chosen to use vrporn it would be vrporn-1044785 (rather than lustreality-1044785).  So, I would still have the possibility of someone sharing scene data with a different id. I'm not sure it really helps a lot in this situation.

Ideally, we would use a global id, something like stashdb.  But I don't believe there is or will ever be a reliable source.  Stashdb comes close, but it doesn't have all scenes especially older ones or more obscure studios.  Even recent scenes for the main sites, can take a few days before someone adds them.

I like that I can tell what studio a scene is for from looking at the scene_id, but the reality is, if I was doing a query, I always use the site column, so your right is doesn't add anything.  If it was just "slr-" or "vrporn-", it would tell you what scraper was used.  There isn't anything in the scene data that will tell you that at the moment. That is probably the biggest advantage to a change.

I don't think existing data dump is an issue, there more like backups, you normally want a fairly recent one anyway. I try to create new one as a backup about once week.  If I need one for testing like the Realjamvr scraper, I create a new one for that.  It may affect other systems that link with xbvr, like timestamp.trade, but the change to Realjamvr's site means everyone is going to have to tackle changing Ids.

I am not sure if there is an advantage doing it here or whether it should be treated as its own separate PR.

crwxaj:
I think preventing duplicate scenes from different scrapers is a lost cause, and should lie in the responsibility of the user. If we have the same studio prefix from different scrapers we might run into even more problems, if the ID ranges of these scrapers overlap. At least you can determine the source, if you have the same scene as slr-123 and povr-9876.

So I would propose:
- Build a migration to change all scene IDs from SLR studios and other aggregator site to "slr-", "povr-", "vrphub-" and "vrporn-" IDs
- remove the studio from the scrapers.json
- Thin out the SLR scraper list a bit (i. e. remove studios with less then 5 releases in the last year)
- Warn the users about the change in the release notes, I might even call the next release 0.4.0 to emphasize the issue

toshski:
I'll make a list of the one's sites I'll think will be dropped.
Can I check what you mean by remove the studio from the scrapers.json. The file only contains studios or are you talking about removing a specific field.  I think I still need them for the site record, even if I'm not using them for the prefix any more

toshski:
The following is a list of studios with less than 5 scenes since 2022-02-01. I would keep SLR JAV Originals and ThatRandomEditor they are both new within the last 6 months

FATP (SLR)	4
ThatRandomEditor (SLR)	4
DANDY (SLR)	3
MaturesVR (SLR)	2
SLR JAV Originals	2
HentaiVR (SLR)	1
LeninaCrowne (SLR)	1
VRFanService (SLR)	1
WAAPVR (SLR)	1
AD4X (SLR)	0
AltPorn4uVR (SLR)	0
Anal Delight (SLR)	0
Babykxtten (SLR)	0
BurningAngelVR (SLR)	0
Cumtoon (SLR)	0
DDFNetworkVR (SLR)	0
GranniesVR (SLR)	0
HerFirstVR (SLR)	0
HoliVR (SLR)	0
Hookfer (SLR)	0
JustVR (SLR)	0
JVRPorn (SLR)	0
LezVR (SLR)	0
MarrionVR (SLR)	0
MMM100 (SLR)	0
Net69VR (SLR)	0
Pornbcn (SLR)	0
PVRStudio (SLR)	0
ScrewBoxVR (SLR)	0
SLR Originals BTS	0
VirtualPee (SLR)	0
VReXtasy (SLR)	0
VRFirstTimer (SLR)	0
VRpussyVision (SLR)	0
xVirtual (SLR)	0

toshski:
In my test environment for the Realjamvr change, I added a legacy_scene_id (and legacy_ scene_url) column, this was just in test so I could make a cross reference list.  I wonder if it is worthwhile adding an legacy_scene_id column to store the original id in case anyone needs it.

May also be worth warning people a couple of days before the release to change their docker address from :latest if they don't want the update straight away.

crwxaj:
> Can I check what you mean by remove the studio from the scrapers.json.

Sorry, I meant to write "remove the studio ID from scrapers.json", since we don't use it for scene_ids anymore, but I guess it is still required for the URL to be scraped. I wonder if it would be better to use the whole URL (e.g. https://www.sexlikereal.com/studios/lustreality) in the scrapers.json to make it more obvious what has to be put there.
Also I wonder if the company field is really required. scene.Studio doesn't seem to be really used anywhere...

> The following is a list of studios with less than 5 scenes since 2022-02-01. I would keep SLR JAV Originals and ThatRandomEditor they are both new within the last 6 months

Also leave FATP and DANDY, since they also have somewhat recent scenes. The rest can go.

> legacy_scene_id

Sounds like a good idea to me

> May also be worth warning people a couple of days before the release to change their docker address from :latest if they don't want the update straight away.

I'm afraid we don't really have an outlet to warn users, besides Discord. But then again, most users should be fine to update.

pkg/config/scraper_list.go
```diff
@@ -0,0 +1,159 @@
...
+	return nil
+}
+
+func RemoveCustomListNowOffical(customSiteList []ScraperConfig, officalSiteList []ScraperConfig) []ScraperConfig {
+	newList := []ScraperConfig{}
+	for _, customSite := range customSiteList {
+		if !CheckMatchingSite(customSite, officalSiteList) {
+			newList = append(newList, customSite)
+		} else {
+			offileSite := GetMatchingSite(customSite, officalSiteList)
```
crwxaj, line 102: Found another typo ;)

toshski:
I do like your idea of using the URL in the scraper list instead of the Id, as it is something that is more meaningful and obvious to the users.  However, the Id is also needed as it is the Id of the Sites table.  So, I have added the URL and made the Id field optional. If the Id is not provided it will default based on the URL. The scrapers will now use the URL provided from the scraper site list rather than building it in code from the Id. This means the Id and URL page don't have to be the same which has a some of advantages, e.g., it works for VRPHub where Stripz and VRHush URLs and ids didn't follow the normal pattern and were essentially hardcoded.

It also means, if someone wants to use a different aggregator site from the official list, maybe they have a sub to the other site. They can now just specify a different Id from the official one and scrape that. You could have done that without the URL, but it would have been tricky setup.  Given they could now have duplicate scenes from different aggregators, I have also added Attribute filters for "POVR Scraper", "SLR Scraper", "VRPHub Scraper" and "VRPornScraper",

I have left the company field but added a note that it is optional.  Likewise, I can't think of anything it is used for.  But it is set in every other scraper. I did check if removing it from the scraper would be all that's required to remove it completely. 
 Unfortunately, it is used in a few old migrations to identify records that needed a change, at that point I decided to leave it.

I have deleted records from the site table, so users don't see the sites where the scraper has been removed.

crwxaj:
>  I have added the URL and made the Id field optional. If the Id is not provided it will default based on the URL.

What use would it have to provide your own ID vs. just using the generated one?

Also, it just came to my mind, that we probably also have to migrate the video previews.

toshski:
I have had problems with the Import/Export functions to work.  If you are selecting "Only enabled sites", it could not find the renamed scene_ids.  It was finding the match site based on the old scene_id prefix.  I fixed it for the Official list, but the Custom list still would not work, I then fixed it for the Custom list and then the Official List would break or even the scenes not changed.  I may have been missing something obvious but seem to keep going in circles.  The difficultly I was having, was because there isn't anything the links the scenes and the site records directly in the DB, you have to write some logic to match and the scene_id code has made that part more complex.  I have had this issue on a few times with my own personal mods and was a problem for #1095 the Site Subscription mod.  In the end, I took the relevant code from that PR, that adds another column on the scene to hold the scrapper id for that scene, which is set in each scraper during the scraping process.

I'll do some more testing first and then can upload the new commits, probably in about 18 hours, so around 2023-02-03 6:00 GMT

crwxaj:
No need to hurry, I'm gone for the weekend. Let's target the release for next Wednesday.

toshski:
Handled the renaming the video previews.

I have added a scraper_id to the scene record.  This holds the Site ID that the scene was scraped from.  The code was taken from the Site Subscription PR #1095 
This proved useful to address the issues for Importing and Exporting bundles. It also simplified the logic required to shift scenes when a Studio shifts from the custom list to the Official list.

I had quite a few additional Studios in my private fork.  I intended to handle these manually as that is the risk of doing your own mods.  A couple of these didn't follow the normal naming, mainly due to Uppercase characters in the url.  This created a lot of duplicates that were hard to identify.  While I don't think many people would add Studios in their own fork, I do expect I'm not the only, s, I have made the migration more robust, to try and handle studios in private forks.  The new Scrapper Id on the scene was very useful here as well.

To avoid users accidentally importing old bundles and importing duplicate scenes with the old scene_ids, I have bumped up the version number for bundles to version 2.1 and the import won't import the previous version 2 exports.  The formats of Versions 2 and 2.1 are still compatible, so if someone wants to, they can manually change the version number in the bundle file and import it.

I have added a bundle export before migrating the Scene Ids as an additional backup step.  This was mainly after seeing the difficulties sorting out the data with duplicate or scenes not migrating for the Studios in a private forks. But it may be a good backup in case something unforeseen happens for any user.  It will export as version 2, so it can only be imported into an earlier release and saved in the app's download directory.  This will add to the time for the migration, although there are plenty of progress messages in the log.  If you don't think we should export a bundle backup, we can remove the following from migrations.go
```
common.Log.Infof("Creating pre-migration backup, please waiit, backups can take some time on a system with a large number of scenes ")
tasks.BackupBundle(true, false, true, true, true, true, true, true, true, true, true, "0", "xbvr-premigration-bundle.json", "2")
common.Log.Infof("Go to download/xbvr-premigration-bundle.json, or http://xxx.xxx.xxx.xxx:9999/download/xbvr-premigration-bundle.json if you need access to the backup")

```

crwxaj:
I had a first test and so far it is looking fine. A few notes:
- I think we can remove SLR Labs as well as it more like a tutorial and test area for SLR
- I think we should move WankzVR and MilfVR to the POVR scraper as well on this occasion.

toshski:
TranzVR as well, it in the wankzvr scraper

PR #1118, opened 2023-02-06
Realjam id using scene url

toshski:
This PR has been prepared in case the Scene_Id for RealjamVR needs to be changed to be based of the Scene's URL. 

Recently, RealjamVR went made a lot of changes to their site and initially, in the first four days, there was no Id that could be used to identify a scene.  After 4 days, the trailer filenames changed to include the existing ID. The rewrite of the scraper to handle the new layout and to use the ID from the trailer filenames has been done in #1117 

However, I am concerned that whether the Trailer names will continue to have a useable Id long term or whether this is only working for the short term for existing scenes, or ones already been prepared for release.

- The PR will log the original Scene_Id to the Legacy_Scene_Id added in #1113 
- Will rename Video Preview files

This PR does not need to be merged unless it became necessary.  I had coded a change to use the Scene URL as a basis of the Id until the existing ID became available again.  But thought I would prepare a separate PR to save time in case it is needed and not lose the code change.  It can also be merged anyway, if it seems a better choice to use the URL rather than the existing Id, even though it is available.


toshski:
The Ids have jumped to a completely new number range, so I expect new scenes since their site refresh have started to come through.  Since using the trailers for the Ids still works with these, I am closing the PR as it does not look like it will be needed.

PR #1120, opened 2023-02-10
hide preview gen exec cmd window on windows os

JustinRoper:
this will allow on windows to hide the cmd prompt from being shown when generating preview.

tested locally on windows and working for me there. took me a min to figure out your build steps to compile but figured it out. using my version locally until we can get this merged into master/released.

need someone to check that the non-windows versions working as expected/normally as well since that is not something i have access to locally at this time.

### DIFF
--- a/pkg/tasks/preview.go
+++ b/pkg/tasks/preview.go
@@ -3,7 +3,6 @@ package tasks
 import (
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


JustinRoper:
I closed issue i added since i created the pr to fix it.

PR #1152, opened 2023-02-23
Tag Groups

toshski:
This feature adds Tag Groups, which are very similar to Actor Aka Groups, except it is for grouping Tags, e.g., you could create Tag group to include, "Slim", "Petite","Slim/Petite", etc

Slight differences from Actor Aka Groups are

- The user specifies the group name
- It does not give warnings if the Tag is used in another Tag Group. This is an expected use case, where for Actor groups it is not.
- Users can rename the group

Aka Actor changes

- Fixed an issue where the Aka Groups list of Actors was not cleared if the group was deleted, leaving orphaned, unreferenced rows
- The speed of removing an Actor from a group has been improved.  The old way was slower and caused timeouts when used for Tag Groups.


crwxaj:
Can I see at any point, which tags are included in a tag group?

toshski:
No easy way to see that.  You can export the Tag groups and check the json bundle or query the database.

toshski:
Added a Tag Lookup button.  If the user has selected a Tag Group (single), they can look up the Tags in the group.  The Tag Filter List will be populated with the Tag Group and a list of the Tags in the Group

PR #1153, opened 2023-02-24
feat: Subtitles support (requires HereSphere v0.8.0)

crwxaj:


pkg/api/heresphere.go
```diff
@@ -533,6 +540,22 @@ func (i HeresphereResource) getHeresphereScene(req *restful.Request, resp *restf
...
+	if err != nil {
+		log.Error(err)
+		return
+	}
+
+	for _, file := range subtitlesFiles {
+		addFeatureTag("Has subtitles")
+		heresphereSubtitlesFiles = append(heresphereSubtitlesFiles, HeresphereSubtitles{
+			Name: file.Filename,
+			URL:  fmt.Sprintf("http://%v/api/dms/file/%v", req.Request.Host, file.ID),
```
Alderamin49ly, line 555: You could fill in the Language from the filename, since subtitles files are usually in the form of `<name>.<language_code>.srt`, with something like:

```
	getLanguage := func(path string) string {
		path = path[:len(path)-4]
		index := strings.LastIndex(path, ".")
		if index > -1 {
			languageCode := path[index+1:]
			languageTag, err := language.Parse(languageCode)
			if err != nil {
				log.Error(err)
				languageTag = language.MustParse("en")
			}
			en := display.English.Languages()
			return en.Name(languageTag)
		} else {
			languageTag := language.MustParse("en")
			en := display.English.Languages()
			return en.Name(languageTag)
		}
	}
```
and then `Language: getLanguage(file.Filename)`

You'd need to add the following imports:
```
	"golang.org/x/text/language"
	"golang.org/x/text/language/display"
```

crwxaj: Thanks, I've added it.

PR #1184, opened 2023-03-19
fix groobyVR Scraper

fucknsfw69:
removing the extra https: added Infront of the scene URL while scraping


theRealKLH:
would you be able to help with the filenames to get the scraper more complete?

theRealKLH:
> removing the extra https: added Infront of the scene URL while scraping

Also remove it from the cover URL...

`		coverURL := "https:" + e.Request.AbsoluteURL(e.ChildAttr("div.player-thumb img", "src"))`

fucknsfw69:
yeah I will work on those. just figured out how to get it build on local. will make the changes and test them. first time working with go

Issue #1043, opened 2022-11-30
Suggestion: Bulk scene creation from folder for unmatched files


Kaylakaze:
Now that we have manual scene creation, it shouldn't be a difficult task to expand that to allow scenes to be generated for a folder, using the file names as scene names. I imagine this would be an option on the Storage tab, a button next to the folders along with rescan. Click and it asks you if you'd like to create scenes for all the unmatched files in the folder, you say yes, and it does, associating the files with each scene as it's created.

gnadenlos:
If you do that, please also add an option to the single unmatched scenes to create a scene from it, because sometimes I have unmatched scenes I don't want to convert to scenes and just wait for the scraper to get the scene  later (for example for studio files where SLR is a few releases behind).

PR #1196, opened 2023-03-24
Adds a Custom Scene for an Unmatched file

toshski:
This feature allows the user to create a custom scene from an unmatched file.

The Title defaults based on the filename, to the same value that is used for the search.
They can create the Scene from just the Title and optionally the Scene Id in a similar way as using Options, Create/Import Scene,Create custom scene.
They can optionally choose to Create the Scene and then Edit the Scene Details.

This should largely meet the requirements of #1043 except it does not do a Bulk create, rather just one by one.  Since the user may not see all files on the screen (due to paging) and Titles generated based on the filename generally require some editing, a bulk create seems potentially dangerous.  A bulk create also seems most useful for an initial setup rather than ongoing use.  Although, extending it to include a bulk create would probably not be hard.


crwxaj:
Could you also add the create&edit feature to Options -> Create custom scene. That's pretty handy.

toshski:
Added to Create Custom Scene

PR #1197, opened 2023-03-24
Change NewValue Column to Text

### LINKED ISSUE #899: Column `actions.new_value` length is too short for SLR Originals scenes in MySQL/MariaDB
This is an extension of https://github.com/xbapps/xbvr/pull/374

That PR extends the size for `new_value` to 4095, which is plenty enough for most studios. However, SLR Originals scene file name arrays are now extremely long. For example, out of the box, the scene `Your Girlfriend, Kira Perez` `scenes.filenames_arr` column is at 5,387 characters. That is fine for the `scenes` table because the `filenames_arr` field is specified as a TEXT field. However, when we match a file with a custom filename to this scene, it will be added to `actions.new_value` and will fail because that field is specified as `varchar(4095)`. This only affects MySQL/MariaDB because SQLite does not enforce input length.

Tagging @SpedNSFW in this because they're the one creating the PR in the first place. Is it reasonable to simply change the type of `new_value` to `text` instead of `varchar`? As far as I can tell we are not querying on that column so changing the field type shouldn't cause any problem.

toshski:
Fixes #899

This changes the new_value column on the actions table from varchar(4095) to Text

I also have very occasionally had issues with updates failing due to the new_value column been too small.  This was only every on non-standard studios that I added to my own build, however, since users can add studios to their own custom list, this may become more of a problem.

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -573,19 +573,19 @@ func Migrate() {
 				var err error
 				switch tx.Dialect().GetName() {
 				case "sqlite3":
-					err = tx.Model(&models.Action{}).Exec("ALTER TABLE actions RENAME TO actions_old").Error
+					err = tx.Model(&models.Action{}).Exec("ALTER TABLE actions RENAME TO actions_old2").Error
 				case "mysql":
-					err = tx.Model(&models.Action{}).Exec("RENAME TABLE actions TO actions_old").Error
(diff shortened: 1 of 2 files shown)


toshski:
Conflict with previous Migrations 0016 0024 occurs when starting a new xbvr system, gets an error with actions_old already exists.  Presumably runs this migration before the delete in migration 0024

PR #1199, opened 2023-03-25
Add search prefix buttons and the ability to search by Date Released or Added

toshski:
This change is in response to requests to make Search Prefixes used to target searching to specific fields, more obvious and ideally easier to use.

The search prefixes introduced in #760 can have a big effect on search results, especially with common words that are ignored by the default search, but a lot of users are not aware they are available.


toshski:
Added the ability to search by the scenes Released or Created (in XBVR) dates by using prefixes released: or added: 

- The default date range is set from a week ago to the current date.
- The default date range does not rule out scenes outside the date range, but it does affect the scoring.
- To excludes scenes outside the date range the user can add a + character, i.e. +released: or +added:

PR #1200, opened 2023-03-25
Fix scraper ID for Czech VR scenes

alexnsfw:
VR Intimacy seems to have been missed in #1113, so their scenes don't have a scraper ID.

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -1401,6 +1401,16 @@ func Migrate() {
 				return nil
 			},
 		},
+		{
+			ID: "0058-add-scraper-id-to-vr-intimacy-scenes",
+			Migrate: func(tx *gorm.DB) error {
+				err := tx.Model(&models.Scene{}).Where("site = 'VR Intimacy'").Update("scraper_id", "czechvrintimacy").Error
+				if err != nil {
+					return err
+				}
+				return nil
+			},
+		},
 	})
 
 	if err := m.Migrate(); err != nil {
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -1411,6 +1411,20 @@ func Migrate() {
 				return nil
 			},
 		},
+		{
+			ID: "0059-fix-czech-vr-scenes-scraper-id",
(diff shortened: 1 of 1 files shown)


theRealKLH:
woouldn't the scaper ID be "czechvr" like all the other CzechVR "sites"? OR should those also have individual scraper IDs?

toshski:
That looks right, the other scraper ids should be czechvrcasting, czechvrfetish, czechvr. I just checked them scraping them on an empty system

alexnsfw:
They all have their own scraper ID; they're defined [here](https://github.com/xbapps/xbvr/blob/master/pkg/scrape/czechvr.go#L163).

theRealKLH:
any idea why all mine are set to "czechvr" then..?

image

theRealKLH:
> They all have their own scraper ID; they're defined [here](https://github.com/xbapps/xbvr/blob/master/pkg/scrape/czechvr.go?rgh-link-date=2023-03-26T18%3A25%3A00Z#L163).

Actually, that's "id" from the "sites" table. "scraper_id" is something new in the scenes table. EVERY "czechvr" based scene in my DB scraper_id = "czechvr" EXCEPT The following....
image

alexnsfw:
The `scraper_id` in the `scenes` table corresponds to an `id` from the `sites` table.

What you're seeing there is another issue with the migration in #1113 that set them all to "czechvr" (because "czechvr" is a prefix of "czechvrcasting" and "czechvrfetish"). Scenes scraped after that migration (i.e. the ones in your last screenshot) have the correct `scraper_id`.

I'll include a fix in this PR.

toshski:
Can you also change the scraper id for Taboo VR Porn, I think these migrated to "taboovr" instead of "taboo-vr-porn"

alexnsfw:
@toshski Those are fine here. Seems to be due to the migration in bb9ac8f, but that was changed before #1113 was merged. I could fix it too, but not sure it's necessary.

toshski:
I had a problem, but that I also had a site "TabooVR" from SLR in my own fork. so it probably messed up the "Taboo VR Porn" scenes on my system.  Thanks for checking

PR #1201, opened 2023-03-25
Update and delete scenes by scraper ID, not site name

alexnsfw:
Some scrapers (e.g. the FuckPassVR ones) share a name, so better use the scraper ID to not e.g. delete more scenes than intended.


crwxaj:
Welcome back, Alex, and thanks for your contributions.

PR #1209, opened 2023-03-29
Option to display Internal Id and HSP Api Json

toshski:
This is a small change designed for Developers & Power Users.
It adds the option to display the Scene's Internal Id and a link to display Scenes Heresphere API Json.

Each can be toggled on/off in Option/Interfaces/Advanced. They are disabled by default.

They are useful for troubleshooting what is JSON is sent to Heresphere or also if you need to access the scene data external to XBVR, e.g. a SQL query tool.

I could have added them to the Web UI settings, but that group was getting a bit full and they would not apply to most users, so I created a new section for Advanced.


toshski:
I can shift it.  I have found it convenient to use from the card list, but given it only an occasional use thing and the clutter aspect, it's probably better on the detail

toshski:
Shifted to detail scene
I think I'll leave it on all the time there.

Funny, I think I saw an XBVR and a Heresphere discord post within 2 hours, that I would have suggested they try this feature to help with diagnose there problem.

PR #1213, opened 2023-04-02
Add Duration Search

toshski:
Added the ability to also search on the Scene Duration.

From the file matching, it will default the search range from 1 less to 1 more than the files duration.

Less useful for the Quick Find as there is no duration to default a range from.  I have defaulted to >=0 so at the users will have an idea of the search syntax to use and can edit it from there.

Sorry I didn't include this in the previous commit, but only thought of it after seeing the duration visible in a number of places to help matching in a recent commit

### DIFF
--- a/pkg/tasks/search.go
+++ b/pkg/tasks/search.go
@@ -26,6 +26,7 @@ type SceneIndexed struct {
 	Id          string    `json:"id"`
 	Released    time.Time `json:"released"`
 	Added       time.Time `json:"added"`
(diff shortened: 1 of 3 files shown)


toshski:
On the assumption this is likely be included in the next release with #1199 I have not added a migration to rebuild the indexes, this would be covered by migration 0060-rebuild-new-indexes in #1199 

If testing in an environment where 0060-rebuild-new-indexes has already run, you will need to manually trigger another index rebuild.

PR #1215, opened 2023-04-05
Display Scene Details using a URL query param

toshski:
This feature you to launch the Scene Detail overlay for a scene using a query parameter (scene_id) in the XBVR URL. e.g. http://localhost:9999/ui/#/?scene_id=badoinkvr-326292, you can also use the scene Internal Id.

I have also created a view (view_scenes) in the database that adds a simple URL (using localhost) for the new param. I usually use DBeaver as my normal SQL tool, it recognizes URLs in a query results, so I can query data, find a scene and just click the row in the URL column to bring up the scene of interest or you can just copy/paste the URL.

This is obviously a feature that will be of any use to a small group, developers and advanced users, people who access the database directly or access XBVR external to the app.   I wasn't sure if it was worth submitting, let me know if not and I'll close & delete the PR

### DIFF
--- a/pkg/api/scenes.go
+++ b/pkg/api/scenes.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-test/deep"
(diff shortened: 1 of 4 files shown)


crwxaj:
I do like the ability to quickly look up a scene_id via URL, but I think your database view is too specific to your use case, and doesn't need to end up everyones database. Anyone who is interested in that should be able to create such a view on their own.

toshski:
No problem, migration removed

PR #1218, opened 2023-04-06
Remove DDFNetworkvr & VRTeenrs Scrapers

toshski:
I have removed the DDFNetworkVr and VRTeenrs scrapers as they no longer work as they are now redirected to completely different sites, Pornbox & Adultprime.

Users wanting scenes can use a Custom Studio to get the scenes metadata, for 
   DDFNetworkVr users can use SLR with all 149 scenes
   VRTeenrs use VRPorn with all 38 scenes or SLR with 37

I have also changed the Options Scraper page to show the Studio name in Red if a Scraper no longer exists and I hide the "Run this Scraper" & "Force update scenes" option under the 3 dots.


vt-idiot:
@toshski I just noticed a bug that's probably related to this? With this commit merged in, if I disable DDFNetworkVR or VRTeenrs after previously having them enabled, the entire list of scrapers turns red, as opposed to just those two. Refreshing the page fixes it, but without doing so, you can't e.g. "Force update scenes" on any scraper since it thinks they're all deprecated.

PR #1251, opened 2023-04-23
Update virtualrealporn.go

### LINKED ISSUE #1250: VirtualRealPorn scraper only scrapes "Tags" but not "Categories" leaving many if not most videos with fairly incomplete tags
Not sure how to go about changing that. They use "categories" for a lot of things you'd expect to be "tags" like positions or acts.

image

```go
		// Tags
		e.ForEach(`a[href*="/tag/"] span`, func(id int, e *colly.HTMLElement) {
			sc.Tags = append(sc.Tags, strings.TrimSpace(e.Text))
		})
		if scraperID == "virtualrealgay" {
			sc.Tags = append(sc.Tags, "Gay")
		}
```

I can't add e.g. 

```go
		e.ForEach(`a[href*="/vr-"] span`, func(id int, e *colly.HTMLElement) {
```

Because that'd add a whole lot of things that aren't those "categories" in as tags.
div class `metaSingleTitle` and `metaSingleData` do only appear to be used for Categories and Tags though...

theRealKLH:
Added Categories to tags. use "Force update" for existing scenes.

This will resolve #1250

### DIFF
--- a/pkg/scrape/virtualrealporn.go
+++ b/pkg/scrape/virtualrealporn.go
@@ -65,7 +65,8 @@ func VirtualRealPornSite(wg *sync.WaitGroup, updateSite bool, knownScenes []stri
 		})
 
 		// Tags
-		e.ForEach(`a[href*="/tag/"] span`, func(id int, e *colly.HTMLElement) {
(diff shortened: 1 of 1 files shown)


vt-idiot (approved):
Works great in testing, including on scenes where there are matching categories and tags.

PR #1270, opened 2023-05-16
Update javdatabase.go

vt-idiot:
`scraper: fix JAV Database cast scraping, add duration`

HTML for scenes on JAVDB changed, "Featured Idols" has been replaced with "ABC-123 Actress/Idols"

seems to work fine in Gitpod


vt-idiot:
Whoops. Probably should've made a separate branch/PR for that, but I figured if I was already poking around in there I might as well add `Duration` to the JAV Database scraper.

PR #1272, opened 2023-05-16
Update heresphere.go

### LINKED ISSUE #1220: Playback from Heresphere doesn't work if there is a percent in the scene title
The URL in the media sources in the Heresphere API is generated with the scene title. If the scene title contains a % symbol, it is not escaped, thus causing an invalid URL that returns a "400 Bad Request" response when called.

This happens with XBVR 0.4.6.

vt-idiot:
Rather than try and sanitize the URL with the full title in it to fix #1220 - I went for a different approach: _drop the title from the URL in the HereSphere JSON entirely_ - and hopefully never have to deal with issues stemming from unescaped **anything** in scene titles ever again.

I don't have any funscripts to test but that's the only other part of the code I touched. `ws` seemingly responds to anything, including the "short" e.g. `http://127.0.0.1:9999/api/dms/file/1234` that this now puts in the HereSphere JSON for `media` -> `url`.

Not sure which change actually fixes this. Not sure if the `file.Filename` that you'd expect to see in the JSON based on Line 224 was never used and the media URLs always had `scene.GetFunscriptTitle()` in them, or if just replacing that with `file.Filename` on Line 338 alone would work, or if `file.Filename` itself was to blame and was just dumping the entire title in there instead of the actual filename. But I'd have to build it again to check, and this already worked just fine as is anyways with seemingly no downsides...`file.ID` is already unique, and `ws` will serve it up without anything appended to it just fine.

Somebody who knows more about how the heresphere API works - ideally whoever wrote heresphere.go - might want to take a look. @toshski @crwxaj

### DIFF
--- a/pkg/api/heresphere.go
+++ b/pkg/api/heresphere.go
@@ -221,7 +221,7 @@ func (i HeresphereResource) getHeresphereFile(req *restful.Request, resp *restfu
 				Height:     height,
 				Width:      width,
 				Size:       file.Size,
-				URL:        fmt.Sprintf("%v://%v/api/dms/file/%v/%v/%v", getProto(req), req.Request.Host, file.ID, file.Filename, dnt),
+				URL:        fmt.Sprintf("%v://%v/api/dms/file/%v/%v", getProto(req), req.Request.Host, file.ID, dnt),
 			},
 		},
 	})
@@ -335,7 +335,7 @@ func (i HeresphereResource) getHeresphereScene(req *restful.Request, resp *restf
 					Height:     height,
 					Width:      width,
 					Size:       file.Size,
-					URL:        fmt.Sprintf("%v://%v/api/dms/file/%v/%v/%v", getProto(req), req.Request.Host, file.ID, scene.GetFunscriptTitle(), dnt),
+					URL:        fmt.Sprintf("%v://%v/api/dms/file/%v/%v", getProto(req), req.Request.Host, file.ID, dnt),
 				},
 			},
 		}


toshski:
I will have to set up some tests, but I think that will break funscripts for people using Script player, Hereshere sends the file name to Script player, which then expects a funscript file with that name.

vt-idiot (commented):
@toshski are you sure funscripts break? Because HSP files and subtitles didn't break - the whole lot of them didn't have the title in the JSON to begin with. The media files were the outlier. Didn't touch the part that puts the Funscript URLs together at all...
https://github.com/xbapps/xbvr/pull/1272/files#diff-a55ad3889a1e1e290ecd5fd8e6997994562bb283337a92e508b85be1b5e62e27L548
Let me know.

toshski:
It looks like it should be ok.  While XBVR and Heresphere both support multiple script files, between Herephere and Scriptplayer, it looks like it only supports one and always uses the Title for the filename.  For scripted scenes the Title sent to Hereshere are changed to "Scene Id - Title", presumably to handle Scenes with the same name.

vt-idiot:
Yeah, perhaps something wasn't working right to begin with?

> Hereshere sends the file name to Script player, which then expects a funscript file with that name.

Take a look at the Heresphere JSON for any scene in 0.4.7 / without my commits. I've never seen the _actual_ filename for a `media` entry in there. Looking at the code, you'd expect to see it, but instead you get the `Scene ID - Title` output in the media URL (the actual "issue" here - `%` is perfectly valid in a scene title - it causes issues in URLs) for `media` entries regardless of whether they have scripts or not.

I'm not quite sure how Scriptplayer works, else I'd offer a suggestion. Do you mean Heresphere only passes one script to Scriptplayer (assuming one `media` entry only - not something like JAV in multiple parts) regardless of how many different `script` entries there are? And is that with or without my commits? Or both?

If it didn't work before anyways, then no big deal. If it did and this broke it, then why not just stick with `Scene ID` for the title that ScriptPlayer sees? Having the scene ID be part of the `media` URL shouldn't cause any issues.

toshski:
The funscript file matched to the scene, isn't actually the file that is used.  You go to Options/Funscripts and Download the funscripts, this creates a zip of all the funscript files, except they are renamed to the format of 'Scene Id  Title.funscript'  You extract the zip file into the directory that is ScriptPlayer is configured to look in.  That's why the 'Scene Id - Title.funscript' filename still matches.

vt-idiot:
Was that a holdover from how it had to be in order to function with the DeoVR API?

toshski:
Probably, it's a long time since I used that.

CoolMintChocolate:
> Was that a holdover from how it had to be in order to function with the DeoVR API?

That was even before the DeoVR API, we used to have to export all the funscript files and copied them to the Quest itself with filenames following that format.
I reverse engineered DeoVR API in order for us to serve funscript dynamically in this PR: https://github.com/xbapps/xbvr/pull/547 In DIscord I asked if we should keep this export functionality, and some people still used it so I didn't take it out as part of that PR.

PR #1273, opened 2023-05-17
scraper: Add VR Queens SLR scraper to config

Wcarr112:


theRealKLH:
@Wcarr112 scrapers.json is different for each person. don't submit PR for your updates of this file.

PR #1276, opened 2023-05-27
Actor Details

toshski:
This modification adds Actor Details and Images to XBVR.  While sounding simple, it is a very large modification and complex in parts due to no way to uniquely identify an Actor and different sites referring to them with different names.  The variability in the availability and quality of actors details is also a complicating factor. 

**Sources of Actors Detail**
- Manual user entry via the UI: This takes precedence over other scraped sources. If a user changes a field containing Actor details, it should not be overwritten by the scrapers. If an item is deleted from a list (Images, URLs, Tattoos, etc.), they should not be re-added.
- XBVR Sites: Actor details will be scraped from XBVR enabled sites where information is available. The breadth and quality of available information vary a lot between sites. If the actor's details (Hair Color, Weight, etc.) are blank and the site scraper has them, they will update the values. When a scene is scraped from a site, the URL to the Actors profile page is recorded and the Actor profiles are scraped after all the scenes have been scraped. This can be disabled and run manually or via a schedule task.  If a site/studio also has an Actor image (not just scene image) on the scene page, this is recorded, however, in most cases the Actor's profile page will need to be scraped for suitable images.
- StashDb: The data quality from Stash is expected to be better than from the XBVR sites. The updates are moderated, and users can correct wrong information. I expect most sites do not update their actor details unless someone complains. StashDb has the most complete set of actor attributes, lacking only a Biography and Weight.  Updates can be run manually or scheduled. To download from StashDb, users need to sign up and get an API Key. StashDb values will overwrite values from XBVR Sites. Matching to StashDb uses strict rules to ensure correct matches. StashDb data may take time to appear and may not include older scenes or smaller studios. The name matching is not reliable, and manual matches can be added by users. Enabling all sites in XBVR results in about 65% of actors being matched to StashDb.

Importing and exporting actor details and links is possible. Importing will not add new actors that are not already in the database. The import can overwrite existing attributes for an Actor or just add missing ones.

**Actor Scraping**
This update includes a scraper and matching process specifically for StashDb. There is also a scraper process for actor details from XBVR sites. The scraper for XBVR sites is a single scraper process for all sites and not separate per site. It utilizes rules defined for each site, relying heavily on CSS selectors, regular expressions (RegEx), and a set of built-in functions (e.g., converting inches to cm, getting country codes, applying RegEx to extract required values, etc.). The rules are defined in code, but these rules can be overridden or added to using the actor_scraper_custom_config.json file in the XBVR directory. The file contains a dummy empty record as a starting point. The built-in coded rules are also exported to the actor_scraper_config_examples.json file, which reflects the live rules. The rules used are a combination of the coded rules and those added to actor_scraper_custom_config.json. 
The actor_scraper_custom_config.json file can potentially be used to create a new scraper for other sites (e.g., ThePornDb or Freeones) by manually entering the actor's URL on those sites and scraping the details, without changing XBVR. However, this may require significant effort for little additional benefit. Modifying the custom json file would require a good knowledge of CSS Selector, Regex Expressions and the ability to read Go code to understand what built-in fuctions are available and what they do and therefore not intended for anyone but advanced users or developers.  
The main use for the actor_scraper_custom file would be to provide updates (via posting to Discord) before a new release is available, that the average user could download or cut/paste.

**StashDB linking**
Matching based on just a matching names would lead to too many errors, I think there was a dozen Actors in Stashdb just called Alexa.  Even factoring in matching Scene for an Actor, there are 20 scenes in XBVR called "My First VR".  XBVR Scenes are first matched with scenes in StashDB based on a matching scene URL, then if the Actor's name or AKA matches an actor in that StasbDb scene will the a link to the StashDB Actor be map.  This will leave Actors unmatched to StashDB that could be. however, I felt it better to have links missing (that can be easily added manually) than to have to fix incorrectly matched Actors.  Scene URLs won't always match, e.g. a scene for MilfVR in XBVR will have a link to POVR and someone could enter into Stashdb the link to the scene on the Wankzvr site, or the Czechvr scenes could uses the czechvr.com link rather than czechvrnetwork.com link.  To handle this additional matching rules can be setup using regex to extract matching identifies from the url or scene id, e.g. you can extract 6325653 from https://povr.com/vr-porn/don-t-wake-daddy-6325653 or https://www.wankzvr.com/milfvr/don-t-wake-daddy-6325653 to match to the scene_id.  Like the scraping rules for Actor Profile pages, these rules can be changed via the custom json file.  While not as complex as the scraper rules, they should only be changed to an advanced user and care must be used to not use too lose a set of rules that may create false matches.

**AKAs**
If you scrape and link details from Stashdb, then possible AKA matches can be suggested.  This is where 2 different Actors in XBVR are linked to the same Actor in Stashdb.  There is a "Possible Aka" actor playlist to filter these. A tab for AKAs will appear in the Actor details if they are part of an AKA group or have possible matches.

**Loading Actor Data for an existing system.**
If a site/studio has actor details that can be scraped, then the URL to the actor's profile page is recorded when scenes are scraped. When upgrading an existing system, there will be no actor details from XBVR sites and scenes will need to be re-scraped to get the URL for the Actors profile pages. Since this can take 4+ hours on a setup with a lot of sites enabled and may not always be desirable, users can import a bundle of Actors details (will post to Discord when included in a new release) to initially populate the Actors details and links to the profiles.  After, importing a bundle the user should still run the Actor Scraping process (Options/Advanced/Actor Setting/Scrape Actor Details).   This should take approx 15-30 minutes.  Assuming the user wants to also scrape details from Stashdb, they need to create an account on Stashdb and then enter the API key into XBVR (Options/Advanced/Actor Details).  This first run of the Stashdb scraping process can take 2 or 3 hours, so they may want to schedule this for a quite period.

Actor bundle attached
[xbvr-content-bundle (actors only).zip](https://github.com/xbapps/xbvr/files/11582968/xbvr-content-bundle.actors.only.zip)


toshski:
There is also a docker available at ghcr.io/toshski/xbvr:latest

toshski:
I've updated to the latest master and addressed Merge Issues with #1277 
With #1280 added to master, I have added an Actor Profile scraping to sexbabesvr
Docker image also updated in case it is useful for testing.

crwxaj:
I had a look and it looks fine to me. Actor data will always be a mess, but this is a good way of handling it. I will merge it in the next release.

PR #1280, opened 2023-06-08
New Sexbabesvr scraper

toshski:
New scraper for Sexbabesvr.

The site no longer has duration for a scene.

Filenames reported by users for downloaded videos don't follow any predictable pattern, sometimes it was the Scene Title, sometimes the Actor name, sometimes a combination, sometimes it was prefixed with sexbabesvr, sometimes just an id and resolution. So I have left the filename list blank. If it settles on an obvious pattern they can be added back. This will mean will have to be matched manually.

The scenes have a release date, but it excludes the year. To handle this, page of scenes (32 scenes) is scraped and a list of all the scenes is made. This is then sorted into order from the newest first, as each page of scenes is not necessarily returned in the order they were called. The scene list is then processed to calculate the year it was released, started with the newest scene and current year, when the month changes from January to December the year for the scene is reduced by 1.

There is no migration required for existing scenes. All scene Uurls have all changed, therefore the next time a user scrapes the site, all scenes will be re-scraped (since those Urls don't exist in the database). When the scene is scraped, the scene Id will match an existing scraped scene in the XBVR database and will update the existing scene with the new Url for the scene page, image, etc, it won't create a new duplicate scene.

### DIFF
--- a/pkg/scrape/sexbabesvr.go
+++ b/pkg/scrape/sexbabesvr.go
@@ -1,17 +1,23 @@
 package scrape
 
 import (
+	"encoding/json"
 	"net/url"
+	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/gocolly/colly/v2"
 	"github.com/mozillazg/go-slugify"
-	"github.com/nleeper/goment"
 	"github.com/thoas/go-funk"
 	"github.com/xbapps/xbvr/pkg/models"
 )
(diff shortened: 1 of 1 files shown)


vt-idiot:
> When the scene is scraped, the scene Id will match an existing scraped scene in the XBVR database and will update the existing scene with the new Url for the scene page, image, etc, it won't create a new duplicate scene.

Will it keep filenames for pre-existing scenes?

PR #1301, opened 2023-06-15
Show scene counts in Actors cards

Alderamin49ly:
As discussed on Discord, this adds the option in the Web UI settings to display the number of scenes and available number of scenes on the Actors cards instead of the scene ratings average and the actor rating.


crwxaj:
I implemented the scene count without the need to choose between scene count and ratings: #1320

PR #1334, opened 2023-07-12
feat: Add ability to scrape a single scene

toshski:
This change allows the user to scrape a single scene using the URL to the scene page from Options/Scrapers, selecting the 3 dots on the scraper to use, has a new option Scrape Single Scene, the user then specifies the URL for the scene.
Uses:

- You can scrape scenes that exist on a site but are missing from the Scene List page used by the scraper.  Sinsvr has had a couple of scenes like this and VirtualPorn recently had all scene from page 5 onwards inaccessible to the scraper
- You don't want to scrape all scenes from a studio, but may see a small subset you still want
- A scene may not be available from a site normally used by your scraper but is available at another site, eg you could use SLR to scrape scenes from JackandJillVR, but it doesn't have the scene "3Some With Nessa", however, the scene is available from VRPorn, by creating an extra Custom Site from JackandJill on the VRPorn site, you could then scrape the single scene "3Some With Nessa" from VRPorn

For some sites the scraper may get some details from the Scene List page as they are not available on the Scene Page.  In these cases the user will be prompted to enter additional details.

Notes

- A scraper does not have to be Enabled to scrape a single scene
- You can specify a url for an existing scene and it will refresh it, similar to the current option to Refresh a scene
- Single scenes cannot be scraped from Baberoticvr as the scraper does not scrape scene pages.

### DIFF
--- a/pkg/api/options.go
+++ b/pkg/api/options.go
@@ -106,6 +106,7 @@ type RequestSaveOptionsPreviews struct {
 type GetStateResponse struct {
 	CurrentState config.ObjectState  `json:"currentState"`
 	Config       config.ObjectConfig `json:"config"`
+	Scrapers     []models.Scraper    `json:"scrapers"`
 }
 
 type GetSearchStateResponse struct {
@@ -590,6 +591,7 @@ func (i ConfigResource) getState(req *restful.Request, resp *restful.Response) {
 
 	out.Config = config.Config
 	out.CurrentState = config.State
+	out.Scrapers = models.GetScrapers()
 
 	resp.WriteHeaderAndEntity(http.StatusOK, out)
 }
--- a/pkg/api/tasks.go
+++ b/pkg/api/tasks.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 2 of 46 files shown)


crwxaj:
Nice feature, but I don't like that you are limited to existing studios. I would like to add a scene from SLR or POVR, without creating a scraper first.

toshski:
I'll have a look, but I'm not sure that it is possible without a much more involved changed.  An instance of a scraper is only created for a Studio not the Site, e.g. "SLR Originals" or "iStripper" not SLR.

toshski:
I have added an option to just specify a URL under Options/Scene Data/Create/Import Scene/Scraper a scene. 
- You can now specify a URL for a Studio from sexlikereal, vrporn, povr or vrphub without having to set up a Custom Site, if it is not in your scraper list.  
- This function will determine the scraper function to call from the URL's domain entered.  Where the domain cannot be used to determine the right scraper function, i.e. any scraper the has multiple Studios, e.g. sexlikereal, vrporn, czechnetwork, etc, then a extra scraper function has been created internally to be used for these URLs and the Studio/Site will be determined from the Scene page, rather than the normal process when registerScraper sets up the scraper callback functions.
- The EditDetails overlay will be shown once the Scene has been scraped, allowing the user to review and enter missing fields as required, e.g. Duration, Release Dateif scraping the scene directly cannot provide those fields. Prior to the latest commits, they were prompted separately, but I think bring up the Edit Details works a lot better.

crwxaj:
Thank you very much. Finally I can fetch the one odd scene for a funscript without scraping the whole studio.

toshski:
Another use for this change is when changes to the Scene List break a scraper but scraping the scene data would sill be fine, i.e. with the recent issue with VirtualTaboo, I was able to scrape recent scene directly before a fix to the scraper was made

PR #1344, opened 2023-07-22
fix VirtualTaboo scraper

theRealKLH:
change in HTML for sceneURL


toshski:
Same change needed for darkroomvr

PR #1354, opened 2023-08-04
Display multiple heatmaps in thumbnail for multipart scenes

Alderamin49ly:
Multipart scenes can have a different funscript for each part, so instead of displaying only the "first" script heatmap in the thumbnail, this changes the thumbnail generation to generate a thumbnail for each funscript (with a limit of 20 heatmaps to avoid completely hiding the thumbnail itself).


theRealKLH:
I'm being lazy. Does it matter if it's NOT a multipart scene? What if it's just 5 different scripts matched to one scene?

Alderamin49ly:
I did it like that do not change the default behavior for regular scenes. I saw that originally it was filtering on is_selected_script so I assumed it was the expected behavior to only display the (first) selected one for regular scenes (though I hadn't noticed before that you could select a script in the UI). But I can remove the check on IsMultipart if that's preferred.

PR #1361, opened 2023-08-25
scraper: add VRSpy

alexnsfw:


theRealKLH:
not adding SLR scrapers any longer. They can be done by each person through the scrapers.json file.

alexnsfw:
I probably should've added a description... :sweat_smile:
But yeah this is a native scraper, for both scenes and actors.

PR #1380, opened 2023-08-31
Handle AI & Multi Axis SLR funscripts

toshski:
Fixes Issue an with SLR AI generated Funscripts no longer flagged as "Has Script Download".

Adds 2 new columns to the scene table to distinguishes between AI generated scripts, Multi Axis scripts from normal scripts on the SLR Site.
Add new Scene Attribute filters "Has AI Script" and "Has Multi Axis Script".


toshski:
Unfortunately, some scenes have both AI and Human generated scripts which SLR just identify as Multi.  So these where missed. 
 But some scenes have 2 human generated scripts and no AI one and they are also Multi.  So unfortunately, I can't just assume Multi are AI and Humun.
 
 Will need to remove this PR for now and rethink how it could be done

toshski:
Closing, need to consider alternative method.

PR #1381, opened 2023-08-31
Fixes Db Locks and Missing SLR Ai Scripts

toshski:
This fixes issues with possible Database locks and scenes with AI Generated scripts not been flagged as having scripts.

Changed from using a gorm Update function which can fail to using the models.Scene Save function which uses a retry.Do and appears to have fixed the problem.

### DIFF
--- a/pkg/api/heresphere.go
+++ b/pkg/api/heresphere.go
@@ -554,6 +554,16 @@ func (i HeresphereResource) getHeresphereScene(req *restful.Request, resp *restf
 			Name: file.Filename,
 			URL:  fmt.Sprintf("%v://%v/api/dms/file/%v", getProto(req), req.Request.Host, file.ID),
 		})
+
+		if scene.HumanScript {
+			addFeatureTag("Hand Crafted Funscript")
+		}
+		if scene.AiScript {
+			addFeatureTag("AI Generated Funscript")
+		}
+		if scene.MultiAxisScript {
+			addFeatureTag("Multi-xis Funscript")
+		}
 	}
 
 	var heresphereSubtitlesFiles []HeresphereSubtitles
--- a/pkg/api/heresphere.go
+++ b/pkg/api/heresphere.go
@@ -561,9 +561,6 @@ func (i HeresphereResource) getHeresphereScene(req *restful.Request, resp *restf
 		if scene.AiScript {
 			addFeatureTag("AI Generated Funscript")
 		}
-		if scene.MultiAxisScript {
-			addFeatureTag("Multi-xis Funscript")
-		}
 	}
 
 	var heresphereSubtitlesFiles []HeresphereSubtitles
--- a/pkg/api/options.go
+++ b/pkg/api/options.go
@@ -65,6 +65,9 @@ type RequestSaveOptionsAdvanced struct {
 	UseImperialEntry      bool   `json:"useImperialEntry"`
 }
 
+type RequestSaveOptionsFunscripts struct {
+	ScrapeFunscripts bool `json:"scrapeFunscripts":`
+}
 type RequestSaveOptionsDLNA struct {
 	Enabled      bool     `json:"enabled"`
 	ServiceName  string   `json:"name"`
@@ -233,6 +236,10 @@ func (i ConfigResource) WebService() *restful.WebService {
 	ws.Route(ws.PUT("/interface/advanced").To(i.saveOptionsAdvanced).
 		Metadata(restfulspec.KeyOpenAPITags, tags))
 
+	// "Web Advanced UI options" section endpoints
+	ws.Route(ws.PUT("/funscripts").To(i.saveOptionsFunscripts).
+		Metadata(restfulspec.KeyOpenAPITags, tags))
+
 	// "Cache" section endpoints
 	ws.Route(ws.DELETE("/cache/reset/{cache}").To(i.resetCache).
 		Param(ws.PathParameter("cache", "Cache to reset - possible choices are `images`, `previews`, and `searchIndex`").DataType("string")).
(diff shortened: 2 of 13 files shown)


crwxaj:
I think it would make sense to mark AI scripted scenes in the database, since SLR dropped a lot of them and I can see the desire to filter them out. Also I think it would be best to add this before we release this PR and we have to migrate existing data afterwards.

toshski:
I've spent most of a day trying to do that. Unfortunately, there seems to be a surprising number of scenes with more than one script.  The scene list page in those cases just says it has Multiple scripts and not if it has AI/Human scripts unless I scrape the scene.  Multiple scripts can also be just 2 Human scripts and no AI Script, so you can't assume Multiple scripts means 1 AI Script + Human scripts.  Given old scenes can have scripts added years later, you would have to continually re-scrape scenes in case they have had scripts added.

The easy solution is to just scrape the scene page of any scene with Multiple scripts to see what scripts it has (scrapes just to script details, not re-scrape the scene data), but that could be a lot more scraping for every scrape run, which I wasn't to keen to do.   So, I decided to release the PR with just the fix for the Deadlock and to at least include the missing scenes that had AI or Multiple scripts that weren't flagged at all, while I continue to look for a nicer solution.  I suspect SLR are going to generate a lot more AI scripts, so it would be really good to be able to filter then separately.  

I'm just building something to scrape all script scenes and dump the data to get a clear idea of the number involved.

Any ideas for alternate approaches?

Not to confuse Multiple scripts with Multi-Axis scripts.

crwxaj:
Keep it simple. Multiple scripts per scene mean there is at least one hand-made script. In my opinion the main purpose of the "AI script" filter is to filter out the plethora of scenes which only have an AI script, so people can avoid them, if they don't like their quality.

toshski:
I've have also considered using the Tags from the api json for the scene details, rather than scraping the scene.  That is should be more efficient than re-scraping the scene but I didn't know if I should start generating a lot of api calls, given SLR pulled the previous api becuase heresphere was using it.  I know we already use it for trailers and passthrough details, but getting that info is a one off.  That is probably the best way, if increasing the use of the api isn't considered a problem.

crwxaj:
I found something. This is a human-only multi-script scene: https://www.sexlikereal.com/scenes/feeling-the-connection-16825?tab=haptic
It has an additional "Sex Toy Script" icon. So my hopothesis would be:
"Sex Toy Script" -> Hand-Crafted only
"Sex Toy Script (AI)" -> AI generated only
"Multi Scripts" -> both
"Multi Scripts" + "Sex Toy Script" -> Hand-crafted only

toshski:
I'll check my data dump and see how many cases of "Multi Scripts" are multiple Human Scripts with no AI Script rather than Human Scripts + AI Scripts.  I know there is some, but I don't know how many.

toshski:
I assume we ignore the Multi-Axis aspect for now, given there is only 4.  
FYI one doesn't show in the list, which is a Multi-Axis AI Generated script  https://www.sexlikereal.com/scenes/guro-gal-creampie-orgy-part-1-12344

toshski:
The number of scenes flagged as Multi Scripts is only about 12.  Much lower than I expected from my initial checks where they seemed more common.  Of these 8 scenes were 2 Hand Crafted scripts and 4 were a Hand Crafted script and an AI script.  Therefore, I have not flagged these scenes as having AI generated scenes as in most cases they don't.  This may can be changed later if the pattern changes.

Scenes with Multi-Axis scripts are flagged when the normal scene scan occurs but chances are the scripts won't exist when scraping shortly after the scene is released.  Hopefully, if they become more common, SLR will make them easier to identify when scanning the Scene List pages

toshski:
Added Feature Tags for Heresphere for AI and Hand Crafted scripts so users can filter them in Heresphere without having to use Saved Seraches from XBVR.

Note: this only applies if the scene has scripts downloaded, using the AI or Handcrafted Script filter in Heresphere will not list scenes that have scripts available in SLR, but the user has not downloaded.

Samsonite121:
I tried this PR because my main PC cannot scrape anymore with version 0.4.12. Every scrape runs into a DB locking and XBVR closes itself. Hardware of this PC is Intel i7-4770 (4 core, 8 threads), 16GB RAM and a Samsung SSD 850 Evo (Sata III). During scraping PC is also at 100% CPU and sometimes mouse cursor hangs shortly. I haven't had this problem the last week on an Gaming Notebook with Intel i7-10750H (6 core, 12 threads), 16GB RAM and M2 SSD. The configuration of sites to scrape (150 sites containing 60 SLR Custom sites) was the same on both machines. I cannot use 0.4.12 with this problem and have to go back to 0.4.11. With 0.4.11 scraping works as usual.

One example log:
time="2023-09-03T19:19:29+02:00" level=info msg="Finished Eve Sweet’s playground (SLR) scraper" completed=true progress=0 scraperID=eve-sweets-playground-slr started=false task=scraperProgress
time="2023-09-03T19:19:33+02:00" level=info msg="visiting https://www.sexlikereal.com/scenes/squirting-waitress-gets-creampie-as-a-tip-36958"
time="2023-09-03T19:19:34+02:00" level=info msg="Finished WankzVR scraper" completed=true progress=0 scraperID=wankzvr started=false task=scraperProgress
time="2023-09-03T19:19:39+02:00" level=info msg="visiting https://www.sexlikereal.com/scenes/virtual-girlfriend-cum-all-over-my-tits-babe-36803"
time="2023-09-03T19:19:41+02:00" level=info msg="visiting https://www.sexlikereal.com/scenes/roommate-creampied-for-rent-36651"
time="2023-09-03T19:19:51+02:00" level=info msg="Finished VRPlayful (SLR) scraper" completed=true progress=0 scraperID=vrplayful-slr started=false task=scraperProgress
time="2023-09-03T19:19:52+02:00" level=info msg="Finished LibertineVR (SLR) scraper" completed=true progress=0 scraperID=libertinevr-slr started=false task=scraperProgress
time="2023-09-03T19:19:56+02:00" level=info msg="visiting https://www.sexlikereal.com/scenes/skinny-latina-fucks-and-moans-like-a-savage-36798"
time="2023-09-03T19:19:56+02:00" level=info msg="Finished YogaVR (SLR) scraper" completed=true progress=0 scraperID=yogavr-slr started=false task=scraperProgress
time="2023-09-03T19:20:02+02:00" level=fatal msg="Failed to save All attempts fail:
#1: database is locked
#2: database is locked
#3: database is locked
#4: database is locked
#5: database is locked
#6: database is locked
#7: database is locked
#8: database is locked
#9: database is locked
#10: database is locked"

toshski:
Just confirming, the problems you are describing is with the PR applied. Also, I assume using SQLite.

toshski:
I've made scraping for Available Funscripts Optional.  Set in "Options/Funscripts/Scrape for Available Funscripts".  To be safe this is off by default, if a user turns it on and has trouble, they know where to disable it.

The updating of the Funscript values is now done using the same Update Channel as for Scraped Scenes.  This should avoid the parallel updates. If an existing scene is found to have Funscripts, the funscript details on the scene are updated, but no other scene details are changed.

crwxaj:
Let me know, when you got these changes ready for release. Also maybe remove the MultiAxis script attribute, as long as it isn't being used.

toshski:
This PR is replaced by #1393 

Done, but I needed to rebase the mod to avoid conflicts generated with other PRs.  But then could no longer Push the Branch back to update the fixes.  So, I have pushed it with a new PR.  

I will close this once the other branch is successfully merged to Master

PR #1409, opened 2023-09-20
Darkroom filenames & Previews attribute

theRealKLH:
* update filenames (requires forced update to change existing scenes)

* add has_preview attribute


theRealKLH:
changed to draft for a hot minute. Maybe a discrepancy in new namings.

PR #1461, opened 2023-10-31
Add option to show heatmaps in the scene list

peter23231:
Adds the funscript heatmaps to the scene list view if a funscript and a heatmap for it is available.
Untitled

Since this makes the list look way more cluttered I made it a toggleable option because not everyone might like it.

Reason for this is I like to have a quick overview of the funscripts I have and which scene/funscript combination looks worth getting without having to go through each detail view individually.  Not sure if this is a usecase for anyone else but I might aswell contribute.


logics102:
Hi there. do you know if there's a way to have this use the heatmap that is selected if they're are more than one? Thank you.

PR #1566, opened 2024-01-04
Added DVD-ID support for the R18dev scraper

pl33x:
- Added DVD-ID support for the R18dev scraper
- Removed message in UI that FANZA content id's are required

### DIFF
--- a/pkg/scrape/r18d.go
+++ b/pkg/scrape/r18d.go
@@ -10,10 +10,10 @@ import (
 	"github.com/xbapps/xbvr/pkg/models"
 )
 
-func getByContentId(req *resty.Request, content_id string) (*resty.Response, error) {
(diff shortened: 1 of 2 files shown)


Alderamin49ly:
I don't think it's a good idea to completely remove the message for content ids. AFAIK R18.dev doesn't have the DVD id for all scenes but always has the Content id. Maybe reformulating the sentence to something like "R18.dev scraper works best with FANZA content ID, e.g. 84vrkm00139, but DVD ID, e.g. VRKM-139, can sometimes work too" would be better.

pl33x:
> I don't think it's a good idea to completely remove the message for content ids. AFAIK R18.dev doesn't have the DVD id for all scenes but always has the Content id. Maybe reformulating the sentence to something like "R18.dev scraper works best with FANZA content ID, e.g. 84vrkm00139, but DVD ID, e.g. VRKM-139, can sometimes work too" would be better.

Updated it

PR #1569, opened 2024-01-07
Update README.md to add command line arguments

diemade:
After spending 1,5 hours searching for the way to change the main port i think that the Command line arguments should be documented here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,4 +134,19 @@ o - previous actor
 p - next actor  
 left arrow - cycles backwards in gallery  
 right arrow - cycles forward in gallery  
-esc - closes details pane    
+esc - closes details pane
+
+#### using Command Line Arguments
+| Command line parameter | Type | Description |
+|------------------------|-------|--------------
+| `--enableLocalStorage` | boolean |Use local folder to store application data|
+|	`--app_dir` | String|path to the application directory|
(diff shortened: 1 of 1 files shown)


toshski:
Can you also add a column for the equivalent Environment Variable, which is more likely to be used by people using Docker. 
XBVR_APPDIR, XBVR_CACHEDIR, XBVR_IMAGEPROXYDIR, XBVR_SEARCHDIR, XBVR_VIDEOPREVIEWDIR,  XBVR_SCRIPTHEATMAPDIR, XBVR_MYFILESDIR, DATABASE_URL, XBVR_WEB_PORT, XBVR_WS_ADDR

Also, as an alternative to keep the Readme shorter, you could add a page to the Wiki and then just reference it in the readme.

diemade:
> 
> Also, as an alternative to keep the Readme shorter, you could add a page to the Wiki and then just reference it in the readme.

There is a wiki?

manacker:
Toshski is referring to the [Wiki feature](https://github.com/xbapps/xbvr/wiki) here on Github, which we only just recently opened for contributions.

crwxaj:
Toshski is referring to the Wiki feature here on Github, which we only just recently opened for contributions.

PR #1573, opened 2024-01-09
Fix  VRHush crashes getting filenames

toshski:
Fixes VRHush crashing for some users.

For some it appears the file attribute in the json data is missing.  This causes XBVR to crash when it is referenced.

I have added checks to ensure it exists before using it. I have also made similar checks to other fields eg title, just in case if they have no value they aren't also missing, rather than existing with a bank value.


toshski:
Fixes #1601

Issue #1570, opened 2024-01-08
PornCornVR/RealJamVR duplicating Actors in a Scene


Asparghus:
When scraping PornCornVR it adds the same actors twice, but with different names. Probably because it has another link on it's description to an actor page.

E.g.: https://porncornvr.com/scene/threesome-passion-girls-prefer-anal/
[Jennifer Mendez](https://porncornvr.com/actor/jennifer-mendez/), [Katy Rose](https://porncornvr.com/actor/katy-rose/) - [Katy Rose VR model](https://porncornvr.com/actor/katy-rose/), [Jennifer Mendez VR celebrity](https://porncornvr.com/actor/jennifer-mendez/) 

https://porncornvr.com/scene/orgy-with-pizza-delivery-guy/
[Jackie Hoff](https://porncornvr.com/actor/jackie-hoff/), [Natalie Brooks](https://porncornvr.com/actor/natalie-brooks/), [Raven Lane](https://porncornvr.com/actor/raven-lane/) - [Jackie Hoff VR star](https://porncornvr.com/actor/jackie-hoff/), [Natalie Brooks VR sweetie](https://porncornvr.com/actor/natalie-brooks/), [Raven Lane VR ideal](https://porncornvr.com/actor/raven-lane/)

EDIT: The same is happening to RealJamVR

https://realjamvr.com/scene/doctors-exam-ava-stone/
[Ava Stone](https://realjamvr.com/actor/ava-stone/) - [Ava Stone VR model](https://realjamvr.com/actor/ava-stone/)

PR #1596, opened 2024-01-18
scraper: Fix duplicate actors on RealJamVR & PornCornVR

theRealKLH:
removes existing duplicates from 12/27/2023 - 01/11/2024. If duplicates newer than these dates, just force update the scene.


theRealKLH:
resolves #1570

PR #1624, opened 2024-02-08
Change SLR CoverUrl to use Link tag

toshski:
Change to the Cover URLs for SLR


vt-idiot:
Why use the HTML at all? #1616

toshski:
covered by #1616

PR #1645, opened 2024-02-16
scraper: Add VirtualPee

awoooooo69:
# Issue
SLR/POVR both have VirtualPee listed, but they don't have all scenes (~45 vs official 150+).
I didn't want to manually populate the metadata for the rest... so

# Solution
I made a new scraper for the virtualpee.com

I threw it together from a few other scrapers and just using css-path from firefox devtools.
Works well enough on my server with my ~100 files.

Sharing it back here in case someone else wants it too.

Cheers

### DIFF
--- a/pkg/scrape/virtualpee.go
+++ b/pkg/scrape/virtualpee.go
@@ -1,0 +1,120 @@
+package scrape
+
+import (
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/gocolly/colly/v2"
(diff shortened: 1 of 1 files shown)


Sunset-1982:
Is it possible to create a scraper for putalocuravr and steelvr from slr, too?

theRealKLH:
> Is it possible to create a scraper for putalocuravr and steelvr from slr, too?

wrong place to be requesting scrapers

PR #1664, opened 2024-03-20
Ability to Limit the Number of Scrapers Running Concurrently

### LINKED ISSUE #1626: Scraper Issue: VRPorn Sites lockdown if high number of requests are detected
I am not sure if slowing down the amount of scrapes will stop triggering the lock down system. But currently big studios like TMW and VirtualReal etc. When trying to scrape the first time will trigger VRPorn to lock you out for 15 mins. When build your linked library it does slow down things having to wait or avoid the big studios till last. I am not sure if its triggered globally or by studio but when I scrape each VRP studio individually I don't trigger it until I hit one of the bigger studios then I get about ~300 scrapes before I get locked and have to wait for 15mins. Obviously launching a larger number of VRPorn Custom Site scrapes at once will also trigger which should also be fixed to limit the number of parallel request sent to VRPorn. (I am unsure if the schedule scrape will also trigger this, as once the library is built the number of scrapers reduces per studio)

toshski:
Fixes #1626 

This allows users to specify the maximum number of scrapers that can run concurrently.  It can be set with the Environment Variable CONCURRENT_SCRAPERS or runtime parameter concurrent_scrapers.

Note: it will run the scrapers in batches, all scrapers in the batch must complete before it will start the next batch, ie if you set the number of scrapers to 10 and 10 are been scraped and one scraper finishes another scraper will not start, all 10 must finish then another 10 will start

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,17 +136,18 @@ left arrow - cycles backwards in gallery
 right arrow - cycles forward in gallery  
 esc - closes details pane
 
-#### using Command Line Arguments
-| Command line parameter | Type | Description |
-|------------------------|-------|--------------
-| `--enableLocalStorage` | boolean |Use local folder to store application data|
-|	`--app_dir` | String|path to the application directory|
-|	`--cache_dir` | String|path to the tempoarary scraper cache directory|
-|	`--imgproxy_dir` | String|path to the imageproxy directory|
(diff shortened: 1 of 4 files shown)


pops64:
Did you mean string or integer for the env variable?

PR #1665, opened 2024-03-21
Ability To Limit the Scraping Rates by Site & Scraper ie Domain

### LINKED ISSUE #1626: Scraper Issue: VRPorn Sites lockdown if high number of requests are detected

toshski:
Fixes #1626 

The change allows you to control the rate at which pages are scraped from a site.  This is most useful when first scraping sites such as Naughty America or VRPorn which have DOS attack preventions.  Although, it may also be useful for day-to-day use if you have a large number of custom sites for VRPorn.

Rate limiting uses the rate limiting options built into colly collectors for most scrapers.  However, this does not work for scrapers that handle multiple sites/studios.  Each studio/site has it's own collector and each would have their own separate rate limits, so in these cases the rate needs to be managed across multiple colly collectors for each studio.  In these cases, the colly Visit methods calls are passed to another method first to manage the flow of calls.  Other scrapers where dealing with multiple studios is not an issue, do not require any changes to implement rate limits, this is managed in the common CreateCollector function they all use

If no rate limits are set, then a scraper performs as it current does.  The Rate Limits are stored in the kvs table, with the key "scraper_rate_limits", the suggested values for Naughty America and VRPorn are:
{ "sites": [ { "name": "www.naughtyamerica.com", "mindelay": 1000, "maxdelay": 2500 }, { "name": "vrporn.com", "mindelay": 4000, "maxdelay": 7500 } ] }

For users not comfortable modifying the database, the config can be loaded via a bundle, they will need to turn on Include Config Settings which is off by default.  Copy of a bundle with suggested values attached.

FYI: Scenes from VRPorn can be scraped faster than the settings allow, however, scraping the Actor details seems to be a lot more sensitive.

People can add other sites if rrequired, these are just the 2 I know of and tested the settings with.
[xbvr-content-bundle (Rate Limits).json](https://github.com/xbapps/xbvr/files/14676024/xbvr-content-bundle.Rate.Limits.json)

### DIFF
--- a/pkg/scrape/genericactorscraper.go
+++ b/pkg/scrape/genericactorscraper.go
@@ -266,10 +266,13 @@ func applyRules(actorPage string, source string, rules models.GenericScraperRule
 			}
 		})
 	}
+	url, _ := url.Parse(actorPage)
 	if rules.IsJson {
-		actorCollector.Request("GET", actorPage, nil, nil, nil)
+		ScraperRateLimiterWait(url.Host)
+		err := actorCollector.Request("GET", actorPage, nil, nil, nil)
+		ScraperRateLimiterCheckErrors(url.Host, err)
 	} else {
-		actorCollector.Visit(actorPage)
+		WaitBeforeVisit(url.Host, actorCollector.Visit, actorPage)
 	}
 	var extref models.ExternalReference
 	var extreflink models.ExternalReferenceLink
--- a/pkg/scrape/povr.go
+++ b/pkg/scrape/povr.go
@@ -126,21 +126,21 @@ func POVR(wg *sync.WaitGroup, updateSite bool, knownScenes []string, out chan<- 
 
 		// If scene exists in database, or the slternate source exists, there's no need to scrape
 		if !funk.ContainsString(knownScenes, sceneURL) && !strings.Contains(sceneURL, "/join") {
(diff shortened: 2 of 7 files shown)


toshski:
I need to refactor this, apparently go maps are not safe for concurrent access from multiple threads.  This can lead to the mutex used not always working and very occasionally may try to unlock an unlocked mutex, which will cause a panic.

toshski:
I have refactored Scraper_Rate_Limits code to use slices instead of a go map, which apparently, is not guaranteed safe for concurrent use.

Issue #1702, opened 2024-04-29
SLR Images are getting 403 error prior to ~4/2020


pops64:
It seems that SLR images being pulled using this format 
`https://cdn-vr.sexlikereal.com/images/13866/85406_desktop.jpg`
 
creates a 403 error for just the cover scene. When requesting an image in this format 
`https://cdn-vr.sexlikereal.com/images/13866/vr-porn-Sex-In-Public-on-the-Balcony-cover-app.webp`

It pulls that format just fine. I am not sure how XBVR is converting the above format to the one it is using. I found the above by inspecting the webpage and looking for the XBVR formatted link and I couldn't find it. It only happens to the cover scene the rest of the gallery images are just fine. I am not sure if SLR has a bug or what. This also may only effect certain studios. Right now it is just the perVRet family of studios on SLR and VRSolos also on SLR, that I have noticed. The above examples come from https://www.sexlikereal.com/scenes/sex-in-public-on-the-balcony-13866

theRealKLH:
going to say you may need to force update your scenes? I have that scene and my image is
[https://cdn-vr.sexlikereal.com/images/13866/vr-porn-Sex-In-Public-on-the-Balcony-cover-app.jpg](url)

notice `.jpg` instead of `.webp`. Also for me that scene is "set for an update", but not updating. 

This scene is from "Lenina Crowne" which was pulled from the built-in list of scrapers. so I'm going to say that is why you aren't updating. You would need to create custom site.

pops64:
My example was freshly pulled from version 24 it effects the underlying SLR scraper for at least a few studios. They made a change to it in #1616

theRealKLH:
I just scraped VRSolos...
![image](https://github.com/xbapps/xbvr/assets/65736720/c4dcba63-4ca9-4f9d-b9b3-7fb00f2cd3af)
What's the problem again?

vt-idiot:
@theRealKLH your scrape of that scene predates the change to how SLR images are scraped. the slug is still in the URL.

the issue occurs because those perVRt scenes and the LC one have comically small cover images. you can see the artifacts/upscaling in the URL that has the scene slug in it. there is no `_desktop` image for them, but there is an `https://cdn-vr.sexlikereal.com/images/13866/85406_app.jpg`

[...]

PR #1721, opened 2024-05-03
Fix for missing cover 403 on low resolution SLR covers

vt-idiot:
Never had to do something like this in Go, if there's a more elegant way than checking HEAD for 200 vs. 403 in a nested if/else like this do let me know.

Tested on:
* the problem scene in #1702
* a currently working scene
* a trans scene


vt-idiot:
Fixes #1702

Issue #1805, opened 2024-07-26
Increase filters request timeout for slow computers (i.e. NAS with a slower HDD)


SR6Sensation:
**Problem:** "api/scene/filters" requests on the homepage usually result in a timeout when I use XBVR on my NAS that only has slower HDDs.

**Effect:** When the filter requests result in a timeout, no filters can be used (i.e. cast, site, tags).

**Solution:** Make filter request timeout configurable or simply increase the timeout to 5 minutes to make sure that the filters can be loaded.

SR6Sensation:
Please add the timeout like this to Filters.vue:
```javascript
    async fetchFilters() {
        this.filteredAttributes=['Loading attributes']
        ky.get('/api/scene/filters', {timeout: 300000}).json().then(data => {
          this.filteredAttributes=data.attributes          
      })      
    }

PR #1806, opened 2024-07-26
fix: Scene filters request timeout (#1805)

SR6Sensation:
Fixes https://github.com/xbapps/xbvr/issues/1805


SR6Sensation:
Fixes #1805

PR #1808, opened 2024-07-30
Fix VRSpy Scene & Actor page changes

toshski:
Updates VRSpy scraper for changes to site layout for both scenes and actor profiles


toshski:
I also changed to only load one actor image, as they all appear to be just slightly different poses, rather than different images

PR #1824, opened 2024-08-20
LethalHardCoreVR Scraper Rebuild for React

pops64:
Fully functionally again. Works 99% of the time. Cards aren't loaded in sequence and my current logic only checks if the last card is loaded before scroll the page to load more. This will occasionally cause a card that is late on loading to be missed as it is scrolled out of range of the viewport. (2 times out of the 10+ times I have ran test) A re scrape usually will pick theses missed ones back up. 

The dockerFile breaks arm docker images. As it uses amd chrome. Probably a way to install arm chrome or may be a non issue. Their appears no way to install chromium on docker Ubuntu that I could find as chromium for Ubuntu's repo throws an error requiring it to be installed via snap.

Functionality that has been removed. Gallery Images aren't scraped. This requires scraping the photos index pages as their doesn't appear any link from the main scene page to the photos. The synopsis as been removed from LH.

Things left todo

- [ ] Trailer - Link is available but I don't understand XBVRs functionally with trailers nor what link to pull. 
- [ ] Docker Arm Compatibility
- [ ] Mac Compatibiltiy
- [ ] WhorecraftVR is no longer available. This should become an SLR custom site for scraping but that would break users DBs. I will print to log Whorecraft no longer available 

Devices I have tested it on
- Windows. Works assuming chrome is installed. Did not try chromium
- Linux. Through yarn dev. Works also requires chromium to be installed
- Docker Tested it on amd64 platform. Works but Ubuntu images requires chrome and not chromium to be installed unless using snap chromium


crwxaj:
I'm sorry, but I won't merge this. I don't want to add that big of a dependency (and maintenance liability) for a single scraper.

PR #1826, opened 2024-08-22
Fix StashDB DB locked errors

pops64:
Since the stashDB scraper is called directly via `scrape.StashDb()` through `api.StashDB()` and not though `tasks.Scrape()` All the locked code that is associated with `tasks.Scrape()` is not used. So DB acitions in `stashDB()` are being blocked. I believe this is causing errors when multiple process attempt to run as other processes lock the DB but stashDB does not and doesn't check either.

Also updated the start code to bring it more in line with how `tasks.Scrape()` starts as `scrapeID` and `siteID` are only used in this one instance.

Testing required to confirm squash

### DIFF
--- a/pkg/scrape/stashdb.go
+++ b/pkg/scrape/stashdb.go
@@ -76,49 +76,53 @@ func StashDb() {
 	if config.Config.Advanced.StashApiKey == "" {
 		return
 	}
-	tlog := log.WithField("task", "scrape")
-	scraperID := "stashdb"
-	siteID := "stashdb"
-	logScrapeStart(scraperID, siteID)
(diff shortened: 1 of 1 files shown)


pops64:
Not squashed yet. I am not sure where the DB is being locked up. Its happening on a `Save` call between `MatchAkaPerformers` and `UpdateAllPerformerData` but all the save calls have identical first prints of error messages ... I will move the lock code higher up the call stack as `StashdbRunAll` is truly the main thread for all stash db functionality

pops64:
I will close this for now and make a new PR once I squash this

PR #1847, opened 2024-09-27
Fix For SexbabesVR Scraper

pops64:
The scene id in on the scene webpage now seems to be 614 for all scenes. Causing all scenes to be rescraped and never adding new scenes.

This pulls the poster url which appears to have a unique identifier in the last directory.

Also updated the cover URL to pull the image used for the thumbnail on the index page. As the latest scene has a SBS image for the cover where the thumbnail contains a more useful image

All appears functional. ~Might require deleting all scraped SexbabesVR scenes due to a shift in scene-id causing file mismatch and scene preview generation hiccups~ Added Migration Code

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -2020,7 +2020,7 @@ func Migrate() {
 				}
 				for _, scene := range scenes {
 
-					if scene.Site == "sexbabesvr" {
+					if scene.Site == "SexBabesVR" {
 						sceneID, err := newSceneId(scene.Site, scene.SceneURL)
 						if err != nil {
 							return err
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/go-resty/resty/v2"
+	"github.com/gocolly/colly/v2"
 	"github.com/jinzhu/gorm"
 	"github.com/markphelps/optional"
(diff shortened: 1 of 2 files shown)


crwxaj:
The scraper update looks fine, but I think we should get rid of the migration. It takes forever without actually changing the scene IDs and in my case it even lead to a duplicate for some reason:
```
...
[dev:go] time="2024-10-10T21:24:10+02:00" level=info msg="Updating sceneid: sexbabesvr-417 to sexbabesvr-417"
[dev:go] time="2024-10-10T21:24:11+02:00" level=info msg="Updating sceneid: sexbabesvr-418 to sexbabesvr-418"
[dev:go] time="2024-10-10T21:24:12+02:00" level=info msg="Updating sceneid: sexbabesvr-419 to sexbabesvr-419"
[dev:go] time="2024-10-10T21:24:13+02:00" level=info msg="Updating sceneid: sexbabesvr-420 to sexbabesvr-421"
[dev:go] time="2024-10-10T21:24:14+02:00" level=info msg="Updating sceneid: sexbabesvr-421 to sexbabesvr-421"
...
```

pops64:
According to KLH there are some scenes that don't have matching scene ids starting around 610. I can add logic to only check scenes starting at 600 and if they match what is already present don't update. This should greatly cut down on the migration time because newer scenes do get a little wonky.

pops64:
Done

PR #1852, opened 2024-10-04
Add BrazzersVR Scraper

pops64:
Appears to currently work. More wide spread testing needed. Had some randomness if valid scene were shown on the index page. The current set of URLs appears to return the correct scenes.

The next index page does not appear to show up in the colly request only on a working browser. Had to resort to checking if any scene links where available and advancing if true.

### DIFF
--- a/pkg/scrape/brazzersvr.go
+++ b/pkg/scrape/brazzersvr.go
@@ -1,0 +1,120 @@
+package scrape
+
+import (
+	"strings"
+	"sync"
+	"strconv"
+	"fmt"
+
+	"github.com/gocolly/colly/v2"
+	"github.com/mozillazg/go-slugify"
+	"github.com/nleeper/goment"
+	"github.com/thoas/go-funk"
+	"github.com/xbapps/xbvr/pkg/models"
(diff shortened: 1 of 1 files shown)


crwxaj:
The scraper works fine for me. However I think the numerical ID from the scene URL would be a better fit for the scene ID than the title slug, since it is less likely to change.
The release date is the same for all scenes, but when I check their date range filter this seems to be the actual case.

pops64:
Yeah. I put in a comment in to code for something to be watched. As bazzers redid their website so who knows if the new scenes will be posted. It does grab the scene ID from the home page url. I tried grabbing everything I could from the json script block as I am indeed worried the class id's will change cause it looks randomized/minifed

pops64:
Update: Moved the bazzers scraper into virtualporn. It shares the same backend API and JSON structure. Everything seems to return correctly. Made adjustments or passed through strings were needed for the differences between sites. Things below are behind the paywall to check


- [ ] Members Link needs to be added for BazzersVR
- [ ] File names need to be checked or skipped over adding them for BazzersVR

PR #1864, opened 2024-10-15
Rebuild of Reality Lovers Fixes #1851

### LINKED ISSUE #1851: RealityLovers scraper broken
It seems that the RealityLovers scraper no longer works.

image

I tried adding https://realitylovers.com/videos/ as a custom site to work around this problem, but the save button is disabled for some reason.

image

What can be done?

pops64:
~API appears to have been abandoned and now returns 403 errors.~ This new scraper now relies on HTML for its data. All works. Title and Cover are both grabbed from index page. These seem to be the most accurate and useful data source. On single scene scrape we fall back to using the gallery and the URL for these sources

### DIFF
--- a/pkg/scrape/realitylovers.go
+++ b/pkg/scrape/realitylovers.go
@@ -1,16 +1,13 @@
 package scrape
 
 import (
-	"fmt"
 	"regexp"
 	"strings"
-	"time"
 
-	"github.com/go-resty/resty/v2"
 	"github.com/gocolly/colly/v2"
 	"github.com/mozillazg/go-slugify"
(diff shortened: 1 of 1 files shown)


pops64:
Closed for now. Will leave the branch there.

pops64:
I am going to leave this open. But the API still works but the code is funky and has logic flaws in it. If wanted I could fix the logic flaws in the OG reality lovers but not really a fan of the 404 status codes and when using curl or a browser to the same address as the scraper you get a access denied message.

pops64:
Per #1851 the current implementation of this is more broken for some users. This should be merged in with the next release

PR #1894, opened 2024-12-28
Stashdb scraper

toshski:
Allows scraping scenes from Stashdb.  Requested a few times on the Discord.  Note: this imports scenes from https://stashdb.org not a local stash database.

You can scrape a single scene, without needing to setup a Studio to scrape. To scrape a whole Studio, setup as a Custom Scrapers.

The Stash Api Key must be setup in Option/Advanced/Actor Settings/Stashdb Api Key.

The scraper is intended for scrapping VR scenes, but does not exclude 2D scenes from been scraped.  Personally, I use and would recommend the Stashdb client for downloaded 2D scenes, but I do also sometimes use the XBVR Stash scraper to download a Studios catalog from Stashdb and peruse with XBVR rather than https://stashdb.org as I feel XBVR has better filtering. I setup a separate instance of XBVR in this case.

The Scene Url will point to the Stashdb Scene, the Scene's Members Url will point to the main Scene Url from Stashdb Scene


Tweeticoats:
stashdb is not the only database, it might be useful to also allow other stash-box instances.
There is https://pmvstash.org/ for porn music videos.
There is https://javstash.org/ and they do have some vr scenes.

toshski:
I won't have time to tackle the extra sites at this stage.  I've got a couple of changes to test and quickly running out of spare time. 
Maybe someone else could or I look at it sometime in the future.

pops64:
This may need manual intervention on acceptance. I also have changed some of the files modified in this PR on #1876. Let me know what needs to be done when the PRs start getting merged in.

PR #1895, opened 2024-12-29
Fix VRSpy Title

toshski:
Fixes missing Title for VRSpy


toshski:
Looks like this is already covered by #1883

PR #1939, opened 2025-01-27
Add arporn.com to scrape

BaBaBaBrap:
Added arporn.com to vrbangers.go scraper


theRealKLH:
you are missing a quote (") after blowvr.com in scenecollector definition (line 21)

PR #1946, opened 2025-02-26
Configure requests and collectors

toshski:
This change allows users to specify request headers, cookies, and the request body. It does not modify headers, cookies, or the body for existing requests made by XBVR—rather, it simply provides users with the ability to set them if they choose to.

This can be useful in scenarios such as bypassing age verification prompts or skipping redirects to advertising pages. The effectiveness of these settings depends entirely on how the site has been implemented.

The configured headers, cookies, and body are stored in a record within the kvs table. When a scraper collector is created or a trailer is requested from the Heresphere API, the system checks whether any modifications need to be applied to the request. For trailers, this can be used to authenticate with the site and access the full scene instead of just the trailer.

The key for the kvs table entry is, by default, the domain of the request followed by either -scraper or -trailers. A UI for configuring these settings is available under Options > Advanced > "Cookies/Headers". Since these details are stored in the kvs table, they can be imported and exported along with bundle imports/exports by selecting "Include config settings". Users can also import individual settings.

Additionally, I have updated the Naughty America trailer functionality to use their Heresphere API instead of a direct URL. This change provides users with more codec and resolution options for trailers.

I will also provide a Wiki page with more complete instructions on how to set this up.

### DIFF
--- a/pkg/api/heresphere.go
+++ b/pkg/api/heresphere.go
@@ -89,12 +89,13 @@ type HeresphereMedia struct {
 	Sources []HeresphereSource `json:"sources"`
 }
 
+type StringOrInt string
 type HeresphereSource struct {
-	Resolution string `json:"resolution"`
-	Height     int    `json:"height"`
-	Width      int    `json:"width"`
-	Size       int64  `json:"size"`
-	URL        string `json:"url"`
+	Resolution StringOrInt `json:"resolution"`
+	Height     int         `json:"height"`
+	Width      int         `json:"width"`
+	Size       int64       `json:"size"`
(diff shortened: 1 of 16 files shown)


pops64:
Should we update the few scrapers that have some cookies hard-coded into the scraper and move them into a default setting kvs. Like with the Custom Scraper json file. I know a few of my scrapers I have added/altered needed a hardcoded age verification cookie to scrape

toshski:
I was going to look at the hardcoded existing scrapers, but after building a UI and the Import functions I didn't have any time left and decided to leave what was working anyway.

PR #2015, opened 2025-06-24
Configurable video extensions, Auto-load more scene cards. New: Gallery Editor. FIX: browser video player not working

Rakly3:
There's a couple of updates rolled into one pull request, but they are sepperated by commit. so they can easily be cherry-picked.

- Update LittleCapriceVR logo URL
- User configurable video extensions with error correction and forbidden etensions (ie .funscript .srt to prevent users from adding them as type 'video')
- QoL scene image Gallery update. Now shows thumbnails instead of a list of URL's. drag-and-drop reordering. Single click set cover. Lockable 'delete image' button (prevent accidental deletion). Error correction, UI/UX only change, the database metadata remains untouched.
- Scenes page now has a toggle to auto load more scenes when scrolling down. When disabled will still show the 'Load More' buttom as before.
- Fixed issue where the webUI video player did not work. (missing draggable and autoload)
- Changes are extensively explained with comments, except for the 'Auto load more' implementation as that code is pretty self-explanatory.

My main concern was making sure that no original behaviour of the app was changed. I would personally change some, but I didn't want to do it as part of the new features' commits.

I hope you'll like the updates as they are high on the users' list of "want to have".

---

Windows binaries - Preview: https://github.com/Rakly3/xbvr/releases/tag/Rakly3-Preview

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
+/.idea
+.DS_Store
+.vscode/
+
+test.db
+/test_cache
+
 /dist/
 
 node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -1,17 +1,57 @@
-/.idea
-.DS_Store
-.vscode/
-
+# Binaries and build artifacts
+xbvr.exe
+.lock
+go.mod
+go.sum
+main.db
 test.db
 /test_cache
 
-/dist/
-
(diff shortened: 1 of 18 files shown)


Rakly3:
this pull request has been split in to smaller pull requests

PR #2045, opened 2025-08-08
Fix missing closing html tags in the UI

benleoau:


Alderamin49ly:
I already submitted a PR to fix the tag https://github.com/xbapps/xbvr/pull/2040
You forgot to fix the next activeTab value here.

benleoau:
you can fix the typo too
Heaaders in

Alderamin49ly:
I'm running xbvr locally on linux and the .air.toml file doesn't seem to be used, maybe you could keep this PR for those change and someone that knows how that's supposed to work can approve it.

PR #2050, opened 2025-08-15
VirtualRealPorn Actor scrape - fix missing image

### LINKED ISSUE #2049: Error 1406: Data too long for column 'image_url' at row 1
XBVR crashed with this error

Image

theRealKLH:
resolves #2049 

scraper fix - actor image was relocated in html code on virtualrealporn so instead of scraping image, the genericactor scraper is scraping json. depending on the size of the json this was (in addition to the image being missing from the Actor's profile) causing xbvr to crash.


benleoau:
the value of image_arr field is wrong too, is this change going to fix image_arr too?
image

PR #2055, opened 2025-09-02
feat: make scene and actor cards cover aspect ratio configurable

Alderamin49ly:
Add option under INTERFACES > Web UI to configure the aspect ratio of the cover images for scenes and actors. And to configure whether the image should fill or fit in the available space:

image


theRealKLH:
Doesn't it currently just depend on which way the actual pic is? One cover could be portrait, the next landscape, etc.? (If I'm understanding this "option") This will force all to be the same. Do we want that?

What will happen here with the option...
image

PR #2093, opened 2025-11-23
Add configurable User Agent option in Advanced settings

belinux:
- Adds a new "Scrapers" tab in Advanced settings that allows users to configure a custom User Agent string
 - Replaces the hardcoded User Agent value throughout the codebase with the configurable option from settings
 - Sets a sensible default User Agent that matches the previously hardcoded value

 ## Changes

   ### Backend
   - **pkg/config/config.go**: Added `ScraperUserAgent` field to Advanced config with default value
   - **pkg/api/options.go**: Added API endpoint support for saving/loading the new User Agent setting
   - **pkg/scrape/scrape.go**: Removed hardcoded `UserAgent` constant
   - **Multiple scraper files**: Updated all scrapers to use `config.Config.Advanced.ScraperUserAgent` instead of the hardcoded value:
     - baberoticavr.go
     - fuckpassvr.go
     - r18d.go
     - slrstudios.go
     - upclosevr.go
     - vrbangers.go
     - vrspy.go
   - **pkg/api/trailers.go**: Updated trailer scraping functions to use configurable User Agent
   - **pkg/migrations/migrations.go**: Updated migration scripts to use configurable User Agent

   ### Frontend
   - **ui/src/views/options/sections/InterfaceAdvanced.vue**: Added new "Scrapers" tab with User Agent input field
   - **ui/src/store/optionsAdvanced.js**: Added state management for the new `scraperUserAgent` setting

   ## Motivation

   This change allows users to customize the User Agent string used by XBVR scrapers, which can be useful for:
   - Working around site-specific User Agent restrictions
   - Testing different User Agent strings without code changes
   - Matching specific browser versions required by certain sites

### DIFF
--- a/pkg/api/options.go
+++ b/pkg/api/options.go
@@ -72,6 +72,7 @@ type RequestSaveOptionsAdvanced struct {
 	ShowHSPApiLink               bool      `json:"showHSPApiLink"`
 	ShowSceneSearchField         bool      `json:"showSceneSearchField"`
 	ScraperProxy                 string    `json:"scraperProxy"`
+	ScraperUserAgent             string    `json:"scraperUserAgent"`
 	StashApiKey                  string    `json:"stashApiKey"`
 	ScrapeActorAfterScene        bool      `json:"scrapeActorAfterScene"`
 	UseImperialEntry             bool      `json:"useImperialEntry"`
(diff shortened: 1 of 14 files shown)


toshski:
Is the User Agent not just another Header, in which case you should be able to do this now under Options/Advanced/Headers-Cookies

belinux:
> Is the User Agent not just another Header, in which case you should be able to do this now under Options/Advanced/Headers-Cookies

Oh! I should have RTFM better the wiki before implementing this, I'll try it later and close the PR if it works as expected, thanks!

PR #2135, opened 2026-01-21
fix: blank Scrapers page on databases with high scene counts

mvdeckard:
## Summary
- Fix blank Scrapers page issue affecting some users with larger scene databases (20k+ scenes) and on resource-constrained systems (NAS/Docker)
- Replace N individual COUNT queries with single GROUP BY query
- Use map-based lookup for HasScraper check instead of nested loop
- Add index on scenes.scraper_id via migration


toshski:
FYI: It probably doesn't matter for adding an index, but migrations that change the schema should go further up the migration list.

Search for the line (currently about 850)
// Put DB Schema migrations above this line and migrations that rely on the updated schema below

PR #2224, opened 2026-07-15
feat: organize files and duplicate analysis

tankxbvr:
Adds an **Organize files** feature plus a **Duplicate files** analyzer.

### Organize
Reorganises library videos into a canonical layout and keeps the `files` table in sync:

```
{TopFolder}/{Studio}/{Studio}.{YY.MM.DD}.{Cast}.{Title}.XXX.{FOV}.{Height}p/
```

- Plans are computed as a **preview** and only applied from the UI. Scheduled runs are **dry-run only** (they log the plan; nothing moves automatically).
- Moves stay **within a storage folder** (auto-detected from local volumes) — files are never moved across folders.
- Byte-identical copies are de-duplicated (md5-confirmed), hard links collapsed, sidecars moved alongside, and emptied directories removed.
- Configurable top folder, a cast-gender preference for folder names, and optional per-actor CamelCase symlink folders.

### Duplicate analysis
Finds scenes with multiple files, flags duration mismatches (possible mis-assignment), compares PSNR between copies, and suggests which to keep. Files can be ignored, deleted, or disassociated from the scene, all from the UI.

### UI
New **Options → Organize files** and **Options → Duplicate files** sections, plus a dry-run schedule under **Task Schedules**.

Self-contained: Go packages build in isolation and the UI (`yarn build`) passes.

### DIFF
--- a/pkg/api/options.go
+++ b/pkg/api/options.go
@@ -186,6 +186,14 @@ type RequestSaveOptionsTaskSchedule struct {
 	LinkScenesHourStart    int  `json:"linkScenesHourStart"`
 	LinkScenesHourEnd      int  `json:"linkScenesHourEnd"`
 	LinkScenesStartDelay   int  `json:"linkScenesStartDelay"`
+
+	OrganizeEnabled      bool `json:"organizeEnabled"`
+	OrganizeHourInterval int  `json:"organizeHourInterval"`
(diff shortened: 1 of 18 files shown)


tankxbvr:
Screenshot 2026-07-16 010320

PR #2233, opened 2026-07-17
scraper: update VirtualRealPorn selectors for new website

ShadowOps:
Website migrated to Laravel Livewire. Key changes:
- Scene ID: script#virtualreal_video-streaming-js-extra -> dl8-video[data-video-id]
- Tags: div.metaSingleData a span -> a.vd-tags__tag
- Cast: JSON-LD actors -> div.vd-pornstar span.vd-pornstar__name
- Duration: MM:SS format -> ISO 8601 PT{N}S (seconds)
- Release date: datePublished -> uploadDate field
- Gallery: figure[itemprop] a -> div.vd-screenshots__grid img
- Listing: videoListContainer/w-portfolio-item-anchor -> a.data-title
- Pagination: ?videoPage=N -> videos/?page=N
- Removed .searchBox option handler (no longer present)
- Download fpName now derived from URL slug

PR #2244, opened 2026-08-08
scraper: update VirtualRealPorn for the new site

DarXV05:
Builds on #2233 by @ShadowOps, which covers the selector migration. Credited via `Co-authored-by`.

## Problem

The site moved to Laravel Livewire and every selector the scraper used is gone. Checked against a live scene page:

| selector | matches |
|---|---|
| `script#virtualreal_video-streaming-js-extra` (scene ID) | 0 |
| `script#virtualreal_download-links-js-extra` (filenames) | 0 |
| `div.metaSingleData a span` (tags) | 0 |
| `figure[itemprop="associatedMedia"] a` (gallery) | 0 |

Scene ID comes from the first, and the scraper ends with `if sc.SceneID != "" { out <- sc }` — so it emits nothing and returns no error. Silent failure.

## Additional fixes beyond #2233

Found while verifying that PR against the live site:

**Performer names carry a `" VR"` suffix** on the new site. Stored verbatim, they create a second actor record alongside the existing one. Stripped.

**Trailers resolve zero sources.** The PR keeps `TrailerType: "scrape_json"` pointed at `trailer.contentUrl` in the JSON-LD, but the `VideoObject` no longer has a `trailer` key at all — its keys are `@context, @type, actor, description, duration, genre, interactionStatistic, name, thumbnailUrl, uploadDate, url`. The stream sources are on the player element instead, which is the shape `badoink.go` already handles: `scrape_html` over `dl8-video source`. Verified to return 5 sources with quality labels.

**Covers silently rejected in some builds.** `image.Decode` validates covers, but `pkg/scrape` registers no image format decoder. It works in the shipped binary only because `pkg/server/heatmapproxy.go` happens to import `image/jpeg` — nothing in `pkg/scrape` guarantees it, and a build excluding `pkg/server` rejects every cover. Registers gif/jpeg/png explicitly.

## Verification

Single-scene scrape against the live site:

```
scene_id   virtualrealporn-<id>    duration  42 min
title      <ok>                    released  2026-02-13
cast       2 (suffix stripped)     tags      8
covers     1                       filenames 25
gallery    7                       trailer   scrape_html / dl8-video source
```

Listing pagination also verified: `/videos/?page=1` yields 20 scene links via `a.data-title`.

### DIFF
--- a/pkg/scrape/virtualrealporn.go
+++ b/pkg/scrape/virtualrealporn.go
@@ -4,15 +4,19 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"html"
 	"image"
+	// image.Decode below only recognises formats whose decoder has been registered.
+	// This package registers none of its own, so cover validation silently rejected
+	// every image except in builds that happened to pull in a decoder elsewhere.
+	_ "image/gif"
+	_ "image/jpeg"
+	_ "image/png"
 	"strconv"
 	"strings"
 
 	"github.com/gocolly/colly/v2"
 	"github.com/mozillazg/go-slugify"
 	"github.com/thoas/go-funk"
-	"github.com/tidwall/gjson"
 	"github.com/xbapps/xbvr/pkg/models"
 )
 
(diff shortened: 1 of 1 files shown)


SturmB:
Independent selector validation against current production HTML on 2026-08-24:

- `the-best-shot-of-the-night` exposes `dl8-video[data-video-id="3150"]`, JSON-LD duration `PT2620S`, upload date `2026-07-13`, 7 `a.vd-screenshots__item` links, 12 `a.vd-tags__tag` links, and 3 performer blocks.
- `between-pages-no-masks` exposes `dl8-video[data-video-id="3161"]` and the expected JSON-LD `VideoObject` metadata.
- `/videos/?page=1` exposes 20 `a.data-title` scene links.

One implementation detail I checked: Livewire replaces the `dl8-video` element after browser hydration, but it is present in the raw server HTML that Colly receives, so the proposed scene-ID selector is appropriate.

I have not deployed this branch, so this is selector-level production validation rather than a compiled end-to-end test. It independently supports the core migration in this PR for the two current scene fixtures above and the listing page.

